PR #3, opened 2021-08-21
Map all shared constants

IotaBread:
Names based on Mojmap. Most of them had `DEBUG_` prefix, I turned some of those into `ENABLE_[...]_DEBUGGER` or just `[...]`.
I have to say, the names order now isn't optimal


OroArmor (approved):

LambdAurora (commented):
Would love to if javadocs with `@value` could be generated for those (only works for primitives).
So in the rendered javadoc there would be the name *and* the value of those constants.

OroArmor:
@ByMartrixx thanks for updating your pr with the new hashed changes!

OroArmor (approved):

LambdAurora (approved):

PR #4, opened 2021-08-21
Rename ThreadedAnvilChunkStorage

Leo40Git:
- `ThreadedAnvilChunkStorage` -> `ThreadedChunkManager`
- `ServerChunkManager.threadedAnvilChunkStorage` -> `ServerChunkManager.delegate`


OroArmor:
What is the reason for removing the `Anvil` word in the names?

Leo40Git:
AFAIK, _all_ of Minecraft's chunk managers use the Anvil format - it's redundant to add it to the class name as well.

Genau6502:
Seems like a pointless change: everyone knows TACS by such, and removing the word anvil seems like it’s completely pointless

OroArmor:
Got it. Wish enigma didn't move stuff around so much

Leo40Git:
@yitzy299: The bigger problem with the name is that it's _not accurate_ - it's not just chunk storage, it's actually a full-blown manager! Therefore, TACS is a bad name, even if it's what everyone is familiar with.

Genau6502:
But if it's a chunk manager you should rename ServerChunkManager

Leo40Git:
I'm not sure if SCM's only purpose is to delegate to a ThreadedChunkManager, therefore I'm leaving its name as-is.

mschae23:
> The bigger problem with the name is that it's _not accurate_ - it's not just chunk storage, it's actually a full-blown manager! Therefore, TACS is a bad name, even if it's what everyone is familiar with.

~~Then use something like `ThreadedAnvilChunkManager`?~~

Edit: It doesn't do the chunk serialization itself, so the chunk format doesn't matter for this class.

Leo40Git:
Why bother keeping Anvil if it's redundant? Even if there's a chunk manager that uses another format, it's in the minority - therefore, it should be named after whatever format it supports while everyone else gets to be "format in name"-free.

TwilightFlower:
Yes, there is no reason to keep Anvil. All the Anvil stuff is in ChunkSerializer iirc.

OroArmor:
It'll be hard to make a proper review of this when half the changes are lines moving. This was something I overlooked and need to fix. I'll won't be able to fix this for around 24 hours, but I'll make it easy for the current prs to update

OroArmor:
@Leo40Git Lets make an issue to discuss this change, and then remake the pull request. I dont think i will be able to easily update the pull request with some cherrypicking as too many lines changed.

Leo40Git:
Done! See #6.

PR #11, opened 2021-09-18
Bunch of 21w37a mappings.

LambdAurora:


mappings/net/minecraft/network/listener/ClientPlayPacketListener.mapping
```diff
@@ -181,6 +185,7 @@ CLASS net/minecraft/unmapped/C_plnddfbr net/minecraft/network/listener/ClientPla
 	METHOD m_wvsleanw onEntitySpawn (Lnet/minecraft/unmapped/C_yvwqrqdq;)V
 		COMMENT Handles the spawning of non-living entities.
 		ARG 1 packet
+			COMMENT the packet
```
mschae23: Was this comment intended?

mappings/net/minecraft/unmapped/C_hvtktbsy.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/unmapped/C_hvtktbsy
```
mschae23: This (and the other new block state providers) should be mapped, the names of these are in `BlockStateProviderType`.

mappings/net/minecraft/unmapped/C_hvtktbsy.mapping
```diff
@@ -0,0 +1,13 @@
...
+	FIELD f_mugtsasn CODEC Lcom/mojang/serialization/Codec;
+	FIELD f_rtvoiawr states Ljava/util/List;
+	METHOD <init> (JLnet/minecraft/unmapped/C_trjftrxr$C_rfufndla;FLjava/util/List;)V
+		ARG 1 seed
+		ARG 5 states
+	METHOD m_dorpdhnl (Ljava/util/List;Lnet/minecraft/unmapped/C_kkdaznyh;D)Lnet/minecraft/unmapped/C_pmeibtxt;
+		ARG 1 states
+		ARG 2 pos
+	METHOD m_nwxwrzvc (Lcom/mojang/serialization/codecs/RecordCodecBuilder$Instance;)Lcom/mojang/datafixers/kinds/App;
+		ARG 0 instance
```
mschae23: This could be something like `fillBlockStateProviderFields`, similar to the method `TrunkPlacer` has.

mappings/net/minecraft/unmapped/C_unqwyjvr.mapping
```diff
@@ -0,0 +1,17 @@
...
+		METHOD m_gdohomuy sampleStartNoise ()V
+		METHOD m_wjebstow sampleNoise ([[DI)V
+			ARG 1 buffer
+			ARG 2 noiseX
+		METHOD m_wpfrryvm sampleEndNoise (I)V
+			ARG 1 x
+	CLASS C_ydritevk
+		METHOD <init> (Lnet/minecraft/unmapped/C_wgwzecpu;II)V
+			ARG 2 x
+			ARG 3 z
```
mschae23, line 17: The rest of the `NoiseInterpolator` mappings (fields and methods) could be added back to this class.

mappings/net/minecraft/util/math/Range.mapping
```diff
@@ -0,0 +1,27 @@
+CLASS net/minecraft/unmapped/C_gbowqdyu net/minecraft/util/math/Range
+	COMMENT Represents a range between two comparable objects from a minimum inclusive to a maximum inclusive.
+	FIELD f_itmrigur min Ljava/lang/Comparable;
+	FIELD f_ptgbnwhf INT_CODEC Lcom/mojang/serialization/Codec;
+	FIELD f_srmndtze max Ljava/lang/Comparable;
+	METHOD <init> (Ljava/lang/Comparable;Ljava/lang/Comparable;)V
+		ARG 1 min
+		ARG 2 max
+	METHOD equals (Ljava/lang/Object;)Z
+		ARG 1 other
```
mschae23: Yarn has the convention to name these parameters `o` (and Quilt Mappings seems to use those as well).

mappings/net/minecraft/util/profiling/jfr/data/CPULoad.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/unmapped/C_acuajsre net/minecraft/util/profiling/jfr/data/CPULoad
```
mschae23:
```suggestion
CLASS net/minecraft/unmapped/C_acuajsre net/minecraft/util/profiling/jfr/data/CpuLoad
```

mappings/net/minecraft/util/profiling/jfr/data/FileIoData.mapping
```diff
@@ -0,0 +1,40 @@
...
+		FIELD f_nxsqxfvf timeSpentInIO Ljava/time/Duration;
+		FIELD f_qiiugwtt totalBytes J
+		FIELD f_ryiojyru topTenContributorsByTotalBytes Ljava/util/List;
+		FIELD f_xkvayfgt countsPerSecond D
+		METHOD <init> (JDJDLjava/time/Duration;Ljava/util/List;)V
+			ARG 1 totalBytes
+			ARG 3 bytesPerSecond
+			ARG 5 counts
+			ARG 7 countsPerSecond
+			ARG 9 timeSpentInIO
```
mschae23:
```suggestion
			ARG 9 timeSpentInIo
```

mappings/net/minecraft/util/profiling/jfr/data/FileIoData.mapping
```diff
@@ -0,0 +1,40 @@
...
+			ARG 10 topTenContributorsByTotalBytes
+		METHOD equals (Ljava/lang/Object;)Z
+			ARG 1 other
+		METHOD m_nopybzuf counts ()J
+		METHOD m_owlefpxw totalBytes ()J
+		METHOD m_rjlvnvro bytesPerSecond ()D
+		METHOD m_snwsxksk countsPerSecond ()D
+		METHOD m_xzdpdluv topTenContributorsByTotalBytes ()Ljava/util/List;
+		METHOD m_yyypxzho timeSpentInIO ()Ljava/time/Duration;
```
mschae23: Here as well

mappings/net/minecraft/util/profiling/jfr/data/GarbageCollectorHeapSummary.mapping
```diff
@@ -0,0 +1,37 @@
...
+		METHOD <init> (Ljava/time/Duration;Ljava/time/Duration;ID)V
+			ARG 1 duration
+			ARG 2 gcTotalDuration
+			ARG 3 totalGCs
+			ARG 4 allocationRateBytesPerSecond
+		METHOD equals (Ljava/lang/Object;)Z
+			ARG 1 other
+		METHOD m_jwnbnzmh getGcOverHeadPercent ()F
+		METHOD m_kkcemshh duration ()Ljava/time/Duration;
+		METHOD m_nrlozewa totalGCs ()I
```
mschae23: Here too, but I'm not sure how that would work with the plural s? Just `totalGcs`?

mappings/net/minecraft/world/biome/source/util/MultiNoiseUtil.mapping
```diff
@@ -0,0 +1,213 @@
...
+			COMMENT from a mathematical point of view, with a special parameter that
+			COMMENT reduces the calculated distance.
+			COMMENT
+			COMMENT <p>For most fields except weight, smaller difference between
+			COMMENT two points' fields will lead to smaller distance. For weight,
+			COMMENT larger differences lead to smaller distance.
+			COMMENT
+			COMMENT <p>This distance is used by the mixed-noise biome layer source. The
+			COMMENT layer source calculates an arbitrary noise point, and selects the
+			COMMENT biome that offers a closest point to its arbitrary point.
```
mschae23: I'm not actually sure how this got into Yarn, but there a few issues with this.

- the special parameter is called `offset`, not `weight`
- it says "difference between two points' fields", but the points aren't compared with each other, only with the actual noise values at the end
- If offset is set to a non-null value, the distance actually gets larger, not smaller.
- It's called "multi-noise biome source", not "mixed-noise biome layer source"

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
```diff
@@ -0,0 +1,122 @@
...
+	FIELD f_ywphpjca humidities [Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;
+	FIELD f_zddmujpv PEAK_START F
+	METHOD m_agfvgbns pickBeachBiome (II)Lnet/minecraft/unmapped/C_hcaelhhx;
+		ARG 1 temperature
+		ARG 2 humidity
+	METHOD m_aluorree addBiomesTo (Ljava/util/function/Consumer;)V
+		ARG 1 biomeEntryConsumer
+	METHOD m_asupgdcz addUndergroundBiomesTo (Ljava/util/function/Consumer;)V
+		ARG 1 parameters
+	METHOD m_awfztaxz addMidSliceTo (Ljava/util/function/Consumer;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;)V
```
mschae23: I'd say you could name these like `addMidBiomesTo` instead of `addMidSliceTo`, so it's at least a bit clearer what they do.

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
```diff
@@ -0,0 +1,122 @@
...
+		ARG 1 temperature
+	METHOD m_ekzzkomy addOffCoastBiomesTo (Ljava/util/function/Consumer;)V
+		ARG 1 parameters
+	METHOD m_ezyzmjyg pickHillBiome (IILnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;)Lnet/minecraft/unmapped/C_hcaelhhx;
+		ARG 1 temperature
+		ARG 2 humidity
+		ARG 3 weirdness
+	METHOD m_fgargpsb getPeaksAndValleysDescription (D)Ljava/lang/String;
+		ARG 0 erosion
+	METHOD m_gaoknfao addUndergroundBiomeTo (Ljava/util/function/Consumer;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;FLnet/minecraft/unmapped/C_hcaelhhx;)V
```
mschae23, line 57: `addCaveBiomeTo`? That would be a simpler name.

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
```diff
@@ -0,0 +1,122 @@
...
+		ARG 2 humidity
+		ARG 3 weirdness
+	METHOD m_unymhsao getNoiseValueDescription (D[Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;)Ljava/lang/String;
+		ARG 0 value
+		ARG 2 parameters
+	METHOD m_vqgtadov pickPlateauBiome (IILnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;)Lnet/minecraft/unmapped/C_hcaelhhx;
+		ARG 1 temperature
+		ARG 2 humidity
+		ARG 3 weirdness
+	METHOD m_xikbagsm addSurfaceBiomeTo (Ljava/util/function/Consumer;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;Lnet/minecraft/unmapped/C_ygdawqho$C_mbbpocmp;FLnet/minecraft/unmapped/C_hcaelhhx;)V
```
mschae23, line 107: Just `addBiomeTo` here? As it's not just for surface biomes, but actually in almost the entire depth range.

mappings/net/minecraft/world/gen/surfacebuilder/NoiseMaterialSurfaceBuilder.mapping
```diff
@@ -0,0 +1,39 @@
+CLASS net/minecraft/unmapped/C_tyauowvy net/minecraft/world/gen/surfacebuilder/NoiseMaterialSurfaceBuilder
```
mschae23, line 1: Not sure if "noise material surface builder" really describes what this class does, it also has the steep slope things.
Is this class used for the stone shores as well? If yes, `AbstractMountainSurfaceBuilder` like Yarn doesn't really fit either.
Not sure.

mschae23 (commented):
Overall, this looks very good (I mostly focused on the world generation parts, though).

mappings/net/minecraft/unmapped/C_hvtktbsy.mapping
mschae23: Oh, I just remembered, there's something weird with them: They're called "Noise 2D", but they're actually 3D.

mappings/net/minecraft/util/math/Range.mapping
LambdAurora: damn it, forgot about it

mappings/net/minecraft/world/gen/surfacebuilder/NoiseMaterialSurfaceBuilder.mapping
LambdAurora: iirc it is used for stone shores.

mappings/net/minecraft/network/listener/ClientPlayPacketListener.mapping
LambdAurora: yes, should have said "the packet to handle" though

mappings/net/minecraft/unmapped/C_hvtktbsy.mapping
LambdAurora: This is a lambda, the one for `RecordCodecBuilder.create()`.

mschae23: Oh. That explains why it's returning an `App` and not a `Product.P...`.
Anyway, the method I actually meant could still be mapped, but it doesn't have to be in this PR.

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
LambdAurora: Not entirely sure, this adds 2 times the biome with depth 0 and depth 1 separately.
Why the underground one is added one time with depth 0.1 -> 0.9

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
LambdAurora: I don't know how much accurate either one is.

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
mschae23: I assume that's because the cave biome one would otherwise be completely inside the surface biome, which would mean either the surface one or the cave one would never generate at that point.

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
mschae23: Both are definitely correct, it's not like one is wrong. "Cave" is just simpler than "underground", and it's only used for cave biomes at the moment.
But this doesn't really matter.

mappings/net/minecraft/block/entity/BlockEntity.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/unmapped/C_hmfagevk net/minecraft/block/entity/BlockEntity
+	COMMENT Represents a block which has a specific instance for every occurences in the world.
```
OroArmor:
```suggestion
	COMMENT Represents an object attached to a certain block which has a different instance for every occurence in the world.
```

mappings/net/minecraft/server/command/JfrCommand.mapping
```diff
@@ -0,0 +1,17 @@
+CLASS net/minecraft/unmapped/C_vkmlkrvv net/minecraft/server/command/JfrCommand
```
OroArmor, line 1: Should jfr be `Jfr` or `JFR`? as it stands for `Java Flight Recorder`, I would lean to `JFR`.

mappings/net/minecraft/unmapped/C_mmhxrdox.mapping
```diff
@@ -1,7 +1,5 @@
 CLASS net/minecraft/unmapped/C_mmhxrdox net/minecraft/unmapped/C_mmhxrdox
-	COMMENT A class holding unknown constants.
-	COMMENT
-	COMMENT @apiNote From the package structure, this class appears to be world-related,
-	COMMENT but it seems to be the exclusive class in a package. Its neighboring packages
-	COMMENT are about portal teleporting and persistent state, yet these constants don't
-	COMMENT seem to relate to those.
+	COMMENT Holds some redstone signal constants.
```
OroArmor: Good. Was wondering when we would get around to this

mappings/net/minecraft/util/profiling/jfr/data/CpuLoad.mapping
```diff
@@ -0,0 +1,15 @@
+CLASS net/minecraft/unmapped/C_acuajsre net/minecraft/util/profiling/jfr/data/CpuLoad
+	FIELD f_cdlhimmg jvm D
```
OroArmor, line 2: Do we want to consider making a new catergory for mapping record compontents? I think its unwise, but it might be useful when mapping, especially when record decompilation is improved, though it might not be hard to make an enigma plugin to find the method that returns the field with ASM and change it via that. @CheaterCodes

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
OroArmor: This also follows the mojmap name

mappings/net/minecraft/world/biome/source/util/OverworldBiomeParameters.mapping
OroArmor: underground is more about intent than cave. this is similar to the mojmap name, which i believe shows that the intent is correct and is clearer to modders what it does.

OroArmor (changes requested):
Looks good to me, nothing major stands out

mappings/net/minecraft/unmapped/C_mmhxrdox.mapping
LambdAurora: I still don't know in which package to put it :pensive:

mappings/net/minecraft/util/profiling/jfr/data/CpuLoad.mapping
LambdAurora: It would be a good thing to have a special category for them.
It would avoid mistypes and it would make it clearer what's a record by reading the file.

mappings/net/minecraft/server/command/JfrCommand.mapping
LambdAurora: [Current conventions](https://github.com/QuiltMC/quilt-mappings/blob/21w37a/CONVENTIONS.md#abbreviations) state that it should be Jfr

mappings/net/minecraft/unmapped/C_mmhxrdox.mapping
OroArmor: Lets put it in the world package

LambdAurora: Sounds cursed to me and a bit mojmap-y but, eh, why not.

OroArmor: It's that or block, which doesn't feel quite right either

PR #21, opened 2021-10-05
Mob and peaceful entities mappings

IotaBread:

PR #31, opened 2021-11-06
Improve some mappings from #21

IotaBread:


Jamalam360 (approved):
These all look like good changes

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
```diff
@@ -23,7 +23,7 @@ CLASS net/minecraft/unmapped/C_mxpeolkc net/minecraft/entity/mob/PiglinBrain
 	FIELD f_pjfnnzxc PLAYER_HOLDING_LIKED_ITEM_MAX_LOOK_DISTANCE I
 	FIELD f_rukeoquj WALK_TO_ITEM_MAX_DISTANCE I
 	FIELD f_sedefqof BARTERING_ITEM Lnet/minecraft/unmapped/C_vorddnax;
-	FIELD f_srhqewhu DESIRED_DISTANCE_FROM_ENTITY_WHEN_AVOIDING I
+	FIELD f_srhqewhu AVOIDING_DESIRED_DISTANCE_FROM_ENTITY I
```
Bubblie01: Should it be AVOID or AVOIDING?

mappings/net/minecraft/entity/passive/FoxEntity.mapping
```diff
@@ -123,16 +123,16 @@ CLASS net/minecraft/unmapped/C_axinhusn net/minecraft/entity/passive/FoxEntity
...
+		FIELD f_cfkermxf EATING_TIME I
 		FIELD f_sojxnnoc timer I
 		METHOD <init> (Lnet/minecraft/unmapped/C_axinhusn;DII)V
 			ARG 2 speed
 			ARG 4 range
 			ARG 5 maxYDifference
 		METHOD m_kqvohfdd pickSweetBerries (Lnet/minecraft/unmapped/C_txtbiemp;)V
 			ARG 1 state
-		METHOD m_olxxahnm pickFromTargetBlock ()V
+		METHOD m_olxxahnm eatSweetBerry ()V
```
Bubblie01: Is this specific to the berry block?

Bubblie01 (commented):
I added two comments that need to be clarified here

mappings/net/minecraft/entity/passive/FoxEntity.mapping
EnnuiL: > 
> 
> Is this specific to the berry block?

They are specific to berries, but not exclusive to sweet berry bushes. Glow berries exist and this goal also picks them from glow vines

Also, I don't understand why the entirely of #15 was silently reverted here; ~~Was it because the original names were on Mojmap?~~ (I misremembered, it wasn't) I think there is room for improvement (In the hindsight, `pickFromTargetBlock` should be something else, like `pickFromTargetPos`, or perhaps something simpler like `attemptBerryPick`?), but as I've mentioned on that PR, the original names don't work well

Bubblie01: I agree with you completely so maybe it should just be eatBerry then?

Bubblie01: Just anything other than eatSweetBerry

LambdAurora: I was going to say that it should be `eatFavoriteFood` since it's also items used for breeding, but turns out that fox's breeding item is actually data-driven with a tag, so that assumption is kinda false then.

So yeah, agreeing with @Bubblie01, `sweet` should be dropped for just berry.

Bubblie01: Pinging me on GitHub smh LOL but yeah changing it to eatBerry makes sense as this is specific to berries not just sweet berry we could also do eatBerries but that might be too confusing

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
LambdAurora: Should be avoid imo: "avoid the desired distance from entity"
it's a constant and do not represent an ongoing action.

LambdAurora (approved):

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
Bubblie01: Agreed

Bubblie01: AVOID_DESIRED_DISTANCE_FROM_ENTITY

mappings/net/minecraft/entity/passive/FoxEntity.mapping
EnnuiL: I still think that `Pick` is a more fitting name than `Eat`; When taking account of similar goals that does have "Eat" in the name and do directly involve, well, consuming food (ex: rabbits with EatCarrotCropGoal and sheep with EatGrassGoal), foxes would be the odd one since unlike others, it doesn't directly involve eating, and I think that would be confusing

This is only my opinion though; If we do stick with `Eat`, then yeah, the "sweet" part should at least be dropped

LambdAurora: Eat is indeed weird since the fox effectively picks berries, then the goal stop, then another goal for actually eating starts later.

Bubblie01: ok pickBerry would be better than in this case I didn't know there was a separate goal for eating

OroArmor (approved):

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
TropheusJ: DESIRED_AVOIDANCE_DISTANCE?

mappings/net/minecraft/entity/passive/FoxEntity.mapping
IotaBread: My bad, I ported some names from yarn and forgot to revert these changes

EnnuiL: Ah, fair enough

IotaBread: Fixed

mappings/net/minecraft/entity/mob/PiglinBrain.mapping
IotaBread: `DESIRED_ENTITY_AVOIDANCE_DISTANCE`?

NoComment1105 (approved):
All seem good to me

Jamalam360 (approved):
Looks good

PR #38, opened 2021-11-13
Port block mappings

IotaBread:


mappings/net/minecraft/block/Block.mapping
```diff
@@ -296,3 +301,6 @@ CLASS net/minecraft/unmapped/C_mmxmpdoq net/minecraft/block/Block
 			ARG 3 facing
 		METHOD equals (Ljava/lang/Object;)Z
 			ARG 1 o
+	CLASS C_znvnetlf
+		METHOD rehash (I)V
+			ARG 1 newN
```
NoComment1105: Is this supposed to have an `N` after it?

NoComment1105 (changes requested):
just the one comment, otherwise good

mappings/net/minecraft/block/Block.mapping
IotaBread: I guess so, as in 'new number'. I should change it though

NoComment1105: Ah I see, something like `newNumber` or `newNo` would work

IotaBread: The method is actually from `Object2ByteLinkedOpenHashMap`, and the parameter name comes from there

NoComment1105: Ah I see. Not something you're thinking of changing then?

IotaBread: I changed it to `newSize`

NoComment1105 (approved):
All good from me

Jamalam360 (approved):

Jamalam360 (approved):
The changes since my last review all look fine :+1:

OroArmor (approved):

PR #52, opened 2021-11-22
Bunch of mappings.

LambdAurora:
Lot to go through, it contains some yarn port, some mojmap somewhere but placed in our structure and checked twice if it made sense.

Have fun.


mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -136,6 +138,8 @@ CLASS net/minecraft/unmapped/C_astfners net/minecraft/entity/Entity
 		ARG 1 boundingBox
 	METHOD m_bqujtsnz isSubmergedInWater ()Z
 		COMMENT Returns whether this entity's hitbox is fully submerged in water.
+		COMMENT
+		COMMENT @return {@code true} if this entity's hitbox is fully submerged in water, otherwise {@code false}
```
IotaBread: You should use a return tag here

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -653,7 +668,10 @@ CLASS net/minecraft/unmapped/C_astfners net/minecraft/entity/Entity
 	METHOD m_slmahvbo getItemsHand ()Ljava/lang/Iterable;
 	METHOD m_smbujbsa discard ()V
 	METHOD m_spboarji isTouchingWater ()Z
-		COMMENT Returns whether this entity's hitbox is touching water fluid.
+		COMMENT Returns whether this entity's hitbox is touching water.
+		COMMENT
+		COMMENT @return {@code true} if this entity's hitbox is touching water, otherwise {@code false}
```
IotaBread: Return tag here as well

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -832,7 +850,7 @@ CLASS net/minecraft/unmapped/C_astfners net/minecraft/entity/Entity
 		ARG 2 yaw
 	METHOD m_yhvbjlmg setPosition (Lnet/minecraft/unmapped/C_vgpupfxx;)V
 		ARG 1 pos
-	METHOD m_ykrmhbpu getEntityWorld ()Lnet/minecraft/unmapped/C_cdctfzbn;
+	METHOD m_ykrmhbpu world ()Lnet/minecraft/unmapped/C_cdctfzbn;
```
IotaBread: This method conflicts with `m_dggmledz`

mappings/net/minecraft/entity/Entity.mapping
LambdAurora: My issue with those tags is it doesn't allow precise return text and brief text.
So either we go detaield with the true if/otherwise false
or we just go with the whether formulation.

mappings/net/minecraft/entity/Entity.mapping
LambdAurora: Urgh, why is Enigma

`m_dggmledz` is supposed to be named `serverWorld` but it forces me to name it the same for no reason?

OroArmor (approved):

mappings/net/minecraft/entity/Entity.mapping
IotaBread: That's odd

IotaBread (approved):

Issue #48, opened 2021-11-18
Create a way to update mappings automatically


OroArmor:
Will likely use a java version of https://gist.github.com/LambdAurora/b987fbb580827566ef1bf52bdde441de

LambdAurora:
My only worry with a Java version of this is stuff like the ASM lib, and stuff to parse tiny/Enigma mappings.

When I looked at those it looked way over-complicated, and possibly-breaking with time (*looks at ASM lib being very strict on bytecode version*).

PR #59, opened 2021-11-28
Add dropInvalidMappings task

IotaBread:
Uses the DropInvalidMappingsCommand I added to Enigma 1.0.5 (QuiltMC/enigma@64e33397429eca91112ee81f6d8c8a73efd37bc5). Would be a part of a solution for #48

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -17,7 +17,6 @@ repositories {
 		name "Mojang"
 		url 'https://libraries.minecraft.net/'
 	}
-	mavenLocal()
 }
 
 dependencies {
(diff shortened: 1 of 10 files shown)


buildSrc/build.gradle
```diff
@@ -17,6 +17,7 @@ repositories {
 		name "Mojang"
 		url 'https://libraries.minecraft.net/'
 	}
+	mavenLocal()
```
OroArmor: Remove before merging

OroArmor (approved):
All good but the mavenLocal()

buildSrc/build.gradle
IotaBread: Done

PR #61, opened 2021-11-30
Screen constants

IotaBread:
Maps most remaining screen constants, and a few other methods

### DIFF
--- a/mappings/net/minecraft/client/gui/screen/Screen.mapping
+++ b/mappings/net/minecraft/client/gui/screen/Screen.mapping
@@ -8,12 +8,16 @@ CLASS net/minecraft/unmapped/C_wrmtlwqx net/minecraft/client/gui/screen/Screen
 	FIELD f_htifmqlv title Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_kmcqeiie width I
 	FIELD f_kreiluct children Ljava/util/List;
+	FIELD f_lpsgbjcj EXTRA_SPACE_AFTER_FIRST_TOOLTIP_LINE I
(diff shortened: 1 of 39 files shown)


LambdAurora (approved):

OroArmor (approved):

mappings/net/minecraft/client/gui/screen/Screen.mapping
```diff
@@ -8,12 +8,16 @@ CLASS net/minecraft/unmapped/C_wrmtlwqx net/minecraft/client/gui/screen/Screen
 	FIELD f_htifmqlv title Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_kmcqeiie width I
 	FIELD f_kreiluct children Ljava/util/List;
+	FIELD f_lpsgbjcj EXTRA_SPACE_AFTER_FIRST_TOOLTIP_LINE I
```
Bubblie01: Maybe take extra out?

Bubblie01 (approved):

mappings/net/minecraft/client/gui/screen/Screen.mapping
Jamalam360: I agree this is quite long and would probably be fine without extra

Jamalam360 (commented):
I sent a question in the internal mappings channel, sorry probably should have added that as a review comment instead :/. I'd imagine it's fine I'm just double checking

Jamalam360 (approved):
The question in the internal mappings channel was resolved. I'm approving now because I'll be busy tomorrow, but Bubblies question is still there.

NoComment1105 (approved):
All good, I agree with Bubblie on his comment, so that's the only thing I've got

PR #94, opened 2022-02-17
Fix broken parameter names in RootSystemFeature

IotaBread:
Fixes broken parameter names in RootSystemFeature.mapping, looks like the parameters shifted because the methods were changed to static


OroArmor (approved):

OroArmor:
Drop invalid mappings should catch this

IotaBread:
Yeah, there is a bug in Enigma I'm trying to fix right now

IotaBread:
The bug has been fixed

LambdAurora (approved):

PR #109, opened 2022-03-03
`util/math/**` mappings

IotaBread:


OroArmor (approved):

LambdAurora (approved):

mappings/net/minecraft/util/math/ChunkPos.mapping
```diff
@@ -1,8 +1,16 @@
 CLASS net/minecraft/unmapped/C_ynrszrtu net/minecraft/util/math/ChunkPos
+	FIELD f_acuruuyv COORD_MASK J
```
triphora, line 2: Is coord an abbreviation we commonly use for coordinate?

triphora (approved):

mappings/net/minecraft/util/math/ChunkPos.mapping
Jamalam360: Not a fan of it personally

NoComment1105 (approved):

mappings/net/minecraft/util/math/noise/InterpolatedNoiseSampler.mapping
```diff
@@ -1,4 +1,27 @@
 CLASS net/minecraft/unmapped/C_uuzbvycz net/minecraft/util/math/noise/InterpolatedNoiseSampler
+	FIELD f_aauagufs cellHeight I
+	FIELD f_boainnds verticalMainScale D
```
supersaiyansubtlety, line 3: I was looking into whether `mainVerticalScale` would make more sense, and noticed something
```java
        this.horizontalScale = 684.412 * config.getXZScale();
        this.verticalScale = 684.412 * config.getYScale();
        this.horizontalMainScale = this.horizontalScale / config.getXZFactor();
        this.verticalMainScale = this.verticalScale / config.getYFactor();
```
are `verticalMainScale` and `horizontalMainScale` always equal to `684.412`, or am I loosing it?
If this is the case, would it make more sense to call them something like `verticalMagicalConstant` and `horizontalMagicalConstant`?

supersaiyansubtlety (commented):
Just one question

mappings/net/minecraft/util/math/ChunkPos.mapping
IotaBread: Coord seems to be the term mojang uses to refer to the conversion to long
image
(0xFFFFFFFFL is `COORD_MASK`, and 32 is `COORD_BITS`)

Any suggestions?

mappings/net/minecraft/util/math/noise/InterpolatedNoiseSampler.mapping
IotaBread: `684.412` probably is a constant defined in another class. Please notice that `horizontalScale` is this constant multiplied by xz**Scale** while `horizontalMainScale` is `horizontalScale` divided by xz**Factor**

supersaiyansubtlety:
```java
this.horizontalScale = 684.412 * config.getXZScale();
this.horizontalMainScale = this.horizontalScale / config.getXZFactor();
```
so if you substitute the expression assigned to `this.horizontalScale` for `this.horizontalScale` in the expression assigned to `this.horizontalMainScale` you get
```java
this.horizontalMainScale = (684.412 * config.getXZScale()) / config.getXZFactor();
```
the multiplication and division cancel out, so it's equivalent (ignoring rounding error) to 
```java
this.horizontalMainScale = 684.412;
```
Since this field is final, it's always equal to `684.412`. 

So would it be appropriate to change the name to reflect this?

triphora: scale != factor

supersaiyansubtlety: oh... my bad

supersaiyansubtlety (approved):

PR #112, opened 2022-03-03
Util & Codecs mappings

IotaBread:


OroArmor (approved):

mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -7,15 +7,41 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
...
+	METHOD m_fsgmnshi (Ljava/util/function/Function;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 element
+	METHOD m_gmlrntec (Ljava/util/function/ToIntFunction;ILjava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 2 element
+	METHOD m_gykcrqhi (Ljava/lang/Integer;)Ljava/lang/String;
+		ARG 0 value
+	METHOD m_ifwybmny (Ljava/util/function/IntFunction;Ljava/lang/Integer;)Lcom/mojang/serialization/DataResult;
+		ARG 1 id
 	METHOD m_ihvqsgya retrieveContext (Ljava/util/function/Function;)Lcom/mojang/serialization/MapCodec;
+		ARG 0 retreiver
```
LambdAurora:
```suggestion
		ARG 0 retriever
```

mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -37,10 +63,19 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
 	METHOD m_kewvhpoe createNonEmptyListChecker ()Ljava/util/function/Function;
 	METHOD m_kfpuczki (Ljava/lang/Object;)Ljava/lang/Object;
 		ARG 0 left
+	METHOD m_lboqbdwi createNonEmptyHolderSetChecker ()Ljava/util/function/Function;
+	METHOD m_lebygrcd rawIdResolverCodec (Ljava/util/function/ToIntFunction;Ljava/util/function/IntFunction;I)Lcom/mojang/serialization/Codec;
```
LambdAurora:
```suggestion
	METHOD m_lebygrcd createRawIdResolverCodec (Ljava/util/function/ToIntFunction;Ljava/util/function/IntFunction;I)Lcom/mojang/serialization/Codec;
```

It feels weird to just have something without really precising what the action is.

mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -54,20 +89,35 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
 		ARG 3 combineFunction
 		ARG 4 leftFunction
 		ARG 5 rightFunction
+	METHOD m_seosjyik (Ljava/util/function/BiFunction;Ljava/util/List;)Lcom/mojang/serialization/DataResult;
+		ARG 1 list
 	METHOD m_sltnwzrc rangedFloat (FFLjava/util/function/Function;)Lcom/mojang/serialization/Codec;
 		ARG 0 min
 		ARG 1 max
 		ARG 2 messageFactory
+	METHOD m_szwqjlsd stringResolverCodec (Ljava/util/function/Function;Ljava/util/function/Function;)Lcom/mojang/serialization/Codec;
```
LambdAurora:
```suggestion
	METHOD m_szwqjlsd createStringResolverCodec (Ljava/util/function/Function;Ljava/util/function/Function;)Lcom/mojang/serialization/Codec;
```

NoComment1105 (approved):

mappings/net/minecraft/util/Util.mapping
```diff
@@ -2,23 +2,29 @@ CLASS net/minecraft/unmapped/C_lvarvugc net/minecraft/util/Util
...
-	METHOD m_awkdanrv toArray (Ljava/util/List;I)Lcom/mojang/serialization/DataResult;
+	METHOD m_awkdanrv fixedSizeList (Ljava/util/List;I)Lcom/mojang/serialization/DataResult;
 		ARG 0 list
 		ARG 1 length
 	METHOD m_cclbyaoh getChoiceTypeInternal (Lcom/mojang/datafixers/DSL$TypeReference;Ljava/lang/String;)Lcom/mojang/datafixers/types/Type;
 		ARG 0 typeReference
 		ARG 1 id
-	METHOD m_drhjlchv toArray (Ljava/util/stream/IntStream;I)Lcom/mojang/serialization/DataResult;
+	METHOD m_chvwiayy getMaxBackgroundThreads ()I
+	METHOD m_drhjlchv fixedSizeList (Ljava/util/stream/IntStream;I)Lcom/mojang/serialization/DataResult;
```
supersaiyansubtlety, line 23:
```suggestion
	METHOD m_drhjlchv array (Ljava/util/stream/IntStream;I)Lcom/mojang/serialization/DataResult;
```
I don't think this should be an overload of `fixedSizeList` considering they return different types.
I'm also a little unclear on the conventions (because new object are wrapped in a `DataResult`), but I think
>Use "create" for methods that create a new instance of some object. Use "get or create" for methods that create a new instance only if one does not already exist. Don't use "get or create" for lazy initialization, though.

applies here, from [CNVENTIONS:Getters, setters, withers, and creators](https://github.com/QuiltMC/quilt-mappings/blob/1.18.2/CONVENTIONS.md#getters-setters-withers-and-creators).

So it may be more appropriate to name them `createFixedSizeList` and `createArray`.

mappings/net/minecraft/util/Util.mapping
```diff
@@ -100,9 +112,21 @@ CLASS net/minecraft/unmapped/C_lvarvugc net/minecraft/util/Util
 	METHOD m_rjgutuoq (Ljava/nio/file/spi/FileSystemProvider;)Z
 		ARG 0 fileSystemProvider
 	METHOD m_rmfqlirq getOperatingSystem ()Lnet/minecraft/unmapped/C_lvarvugc$C_fqjbnqwz;
+	METHOD m_royqyqxg backupAndReplace (Ljava/nio/file/Path;Ljava/nio/file/Path;Ljava/nio/file/Path;Z)V
+		ARG 0 current
+		ARG 1 newPath
+		ARG 2 backup
+		ARG 3 noRestoreOnFail
```
supersaiyansubtlety:
```suggestion
		ARG 3 preventRestoreOnFail
```
to comply with
> Boolean field and variable names should always be adjective phrases or present tense verb phrases

from [CONVENTIONS:General](https://github.com/QuiltMC/quilt-mappings/blob/1.18.2/CONVENTIONS.md#general).

Also consider
```suggestion
		ARG 3 discardOnFail
```
to avoid the double negative?

mappings/net/minecraft/util/Util.mapping
```diff
@@ -100,9 +112,21 @@ CLASS net/minecraft/unmapped/C_lvarvugc net/minecraft/util/Util
...
+	METHOD m_rucymwoq (Ljava/lang/String;)V
+		ARG 0 string
 	METHOD m_skylxoye debugRunnable (Ljava/lang/Runnable;Ljava/util/function/Supplier;)Ljava/lang/Runnable;
 		ARG 0 runnable
 		ARG 1 messageSupplier
+	METHOD m_tcuyutyn backupAndReplace (Ljava/io/File;Ljava/io/File;Ljava/io/File;Z)V
+		ARG 0 current
+		ARG 1 newFile
+		ARG 2 backup
+		ARG 3 noRestoreOnFail
```
supersaiyansubtlety:
```suggestion
		ARG 3 preventRestoreOnFail
```
or
```suggestion
		ARG 3 discardOnFail
```
(see other comment)

mappings/net/minecraft/util/collection/Int2ObjectBiMap.mapping
```diff
@@ -1,12 +1,21 @@
...
 	FIELD f_lgvkwzhs size I
 	FIELD f_poqcloyh values [Ljava/lang/Object;
+	FIELD f_rfzzrczr NOT_FOUND I
 	FIELD f_yyvhlamv nextId I
 	METHOD <init> (I)V
 		ARG 1 size
+	METHOD <init> ([Ljava/lang/Object;[I[Ljava/lang/Object;II)V
+		ARG 1 values
+		ARG 2 ids
+		ARG 3 idToValues
```
supersaiyansubtlety:
```suggestion
		ARG 3 valuesByIds
```
to comply with
> When it's enough, name maps based on the value type. Otherwise, name it in the "valuesByKeys" format.

from [CONVENTIONS:Collections](https://github.com/QuiltMC/quilt-mappings/blob/1.18.2/CONVENTIONS.md#collections)

mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -115,7 +204,19 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
 			ARG 0 pair
 		METHOD m_mdizponh (Lcom/mojang/datafixers/util/Pair;)Lcom/mojang/datafixers/util/Pair;
 			ARG 0 pair
+		METHOD m_oktodooj (Lcom/mojang/datafixers/util/Pair;Lcom/mojang/datafixers/util/Pair;)Lcom/mojang/datafixers/util/Pair;
+			ARG 0 pair
```
supersaiyansubtlety:
```suggestion
			ARG 0 pair1
```
?

supersaiyansubtlety (changes requested):

mappings/net/minecraft/util/Util.mapping
IotaBread: The methods do nearly the same, and the previous name (`toArray`) was just wrong. Maybe I could rename it to `fixedSizeArray` or `fixedSizeStream`
image

About the prefix, I don't think `create` fits here, since it just checks that the provided list/stream has the specified size, and doesn't create any new instance.

mappings/net/minecraft/util/dynamic/Codecs.mapping
IotaBread: Also added prefixes to some of the other methods

mappings/net/minecraft/util/Util.mapping
supersaiyansubtlety: Actually the method whose generic is `int[]` creates a new instance, but the `List` one wraps the passed list. 
Also the `int[]` method is limited to `int`s, so I think `createIntArray` would be appropriate. 

And maybe `verifyListSize` would be appropriate for the `List<T>` method?

supersaiyansubtlety (approved):

mappings/net/minecraft/util/Util.mapping
IotaBread: Keeping the names similar is better imo, since they do (almost) the same

Issue #26, opened 2021-10-31
The name of ReadOnlyChunk is wrong


IotaBread:
Field `f_iichgpub` is named `allowWrites` in MojMap, implying the class does allow writing. The class name in MojMap isn't very descriptive (`ImposterProtoChunk`)

PR #126, opened 2022-03-13
n/m/world/chunk mappings

IotaBread:


OroArmor (approved):

mappings/net/minecraft/world/chunk/ReadOnlyChunk.mapping
```diff
@@ -1,8 +1,9 @@
 CLASS net/minecraft/unmapped/C_riffdstx net/minecraft/world/chunk/ReadOnlyChunk
```
IotaBread, line 1: (See #26) any suggestions?

NoComment1105: Given the fact that it is currently named `ReadOnlyChunk` I'd like to hope there is some grounds to this. As said in #26 it can be a little confusing. To me this says that chunks like this can "be locked" so they're ReadOnly, or "unlocked" so they can be written to. Something around this concept perhaps?

NoComment1105 (approved):

mappings/net/minecraft/world/chunk/palette/PalettedContainer.mapping
```diff
@@ -1,18 +1,52 @@
...
+	METHOD m_eegdzoxi (Lit/unimi/dsi/fastutil/ints/Int2IntOpenHashMap;I)V
+		ARG 1 index
+	METHOD m_faqwumtl getCompatibleData (Lnet/minecraft/unmapped/C_aehiirzw$C_zdxjpjcl;I)Lnet/minecraft/unmapped/C_aehiirzw$C_zdxjpjcl;
+		ARG 1 data
+	METHOD m_haehqoyx (Ljava/util/function/Consumer;Lnet/minecraft/unmapped/C_peoxnqbd;I)V
+		ARG 2 id
 	METHOD m_huyuznrj hasAny (Ljava/util/function/Predicate;)Z
-	METHOD m_iyxhoppi set (IIILjava/lang/Object;)Ljava/lang/Object;
+		ARG 1 predicate
+	METHOD m_iyxhoppi setAndGetUnsafe (IIILjava/lang/Object;)Ljava/lang/Object;
```
supersaiyansubtlety:
```suggestion
	METHOD m_iyxhoppi setUnsafe (IIILjava/lang/Object;)Ljava/lang/Object;
```
Omitting `AndGet` is more in line with Java's `List::set` and `Map::put`.

supersaiyansubtlety (changes requested):

supersaiyansubtlety (approved):

mappings/net/minecraft/world/chunk/ReadOnlyChunk.mapping
NoahvdAa: Perhaps `(Write)LockableChunk`?

NoahvdAa (approved):

PR #138, opened 2022-03-29
Fix #92

### LINKED ISSUE #92: Rename `Fluid#isStill` and associated methods
When we get around to diverging Yarn from upstream, `Fluid#isStill` and all the other `still` references should be changed to `Fluid#isSource`, since it has nothing to do with whether or not the fluid is flowing, but rather whether or not it is a source block. `Fluid#getFlowing` can likely stay the same, but a more accurate name would be nice.

NoahvdAa:
Fixes #92. Not entirely sure whether `m_agslpkrg` should be simplified to Mojang's name, `canExistIn`.

### DIFF
--- a/mappings/net/minecraft/block/BubbleColumnBlock.mapping
+++ b/mappings/net/minecraft/block/BubbleColumnBlock.mapping
@@ -1,7 +1,7 @@
 CLASS net/minecraft/unmapped/C_eopzqcts net/minecraft/block/BubbleColumnBlock
 	FIELD f_jarlnoge SCHEDULED_TICK_DELAY I
 	FIELD f_oylcoxnk DRAG Lnet/minecraft/unmapped/C_xhwijdsd;
-	METHOD m_agslpkrg canBubbleColumnExistIn (Lnet/minecraft/unmapped/C_txtbiemp;)Z
+	METHOD m_agslpkrg canBeBubbleColumn (Lnet/minecraft/unmapped/C_txtbiemp;)Z
 		ARG 0 state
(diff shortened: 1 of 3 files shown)


mappings/net/minecraft/block/BubbleColumnBlock.mapping
```diff
@@ -1,7 +1,7 @@
 CLASS net/minecraft/unmapped/C_eopzqcts net/minecraft/block/BubbleColumnBlock
 	FIELD f_jarlnoge SCHEDULED_TICK_DELAY I
 	FIELD f_oylcoxnk DRAG Lnet/minecraft/unmapped/C_xhwijdsd;
-	METHOD m_agslpkrg isStillWater (Lnet/minecraft/unmapped/C_txtbiemp;)Z
+	METHOD m_agslpkrg canBubbleColumnExistIn (Lnet/minecraft/unmapped/C_txtbiemp;)Z
```
NoComment1105: I don't particularly liek this, perhaps `canBeBubbleColumn`?

NoComment1105 (changes requested):

mappings/net/minecraft/block/BubbleColumnBlock.mapping
NoahvdAa: That sounds a lot better :D

NoComment1105 (approved):

LambdAurora (approved):

NoComment1105 (approved):

IotaBread (approved):

triphora (approved):

PR #144, opened 2022-04-13
New Random System

OroArmor:


mappings/net/minecraft/util/random/PositionalRandomFactory.mapping
```diff
@@ -0,0 +1,13 @@
+CLASS net/minecraft/unmapped/C_coqzjjrt net/minecraft/util/random/PositionalRandomFactory
+	METHOD m_adgvioqq create (Lnet/minecraft/unmapped/C_ncpywfca;)Lnet/minecraft/unmapped/C_rlomrsco;
+		ARG 1 identifier
```
IotaBread: Wouldn't just `id` be better here?

mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -0,0 +1,26 @@
...
+	METHOD m_fakwfybt nextDouble ()D
+	METHOD m_ifdmmgky nextLong ()J
+	METHOD m_imwpekuz createUnsafe ()Lnet/minecraft/unmapped/C_rlomrsco;
+	METHOD m_iwfdczgh setSeed (J)V
+		ARG 1 seed
+	METHOD m_lzlwtgyn nextGaussian ()D
+	METHOD m_mkdxraps nextBoolean ()Z
+	METHOD m_ngqtmirh createLegacy (J)Lnet/minecraft/unmapped/C_rlomrsco;
+		ARG 0 seed
+	METHOD m_paekwdby derive ()Lnet/minecraft/unmapped/C_rlomrsco;
```
IotaBread, line 21: What does this do?

IotaBread (approved):

mappings/net/minecraft/util/random/RandomGenerator.mapping
OroArmor: this mapping already existed, git didnt catch the file rename

IotaBread: Ah

mappings/net/minecraft/util/random/PositionalRandomFactory.mapping
OroArmor:
```suggestion
		ARG 1 id
```

mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -0,0 +1,26 @@
+CLASS net/minecraft/unmapped/C_rlomrsco net/minecraft/util/random/RandomGenerator
```
LambdAurora, line 1: I believe this is the abstracted interface/abstract class.

Would be nice to have some quick javadoc on the class itself.

LambdAurora (approved):

mappings/net/minecraft/util/random/RandomGenerator.mapping
OroArmor:
```suggestion
CLASS net/minecraft/unmapped/C_rlomrsco net/minecraft/util/random/RandomGenerator
	COMMENT An interface describing random behavior in Minecraft.
	COMMENT <p>
	COMMENT This interface should be used instead of {@link java.util.Random} for random number generation.
```

LambdAurora: third line of the diff should be
`	COMMENT <p>`

OroArmor: look good now?

LambdAurora: yup, I guess

mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -0,0 +1,26 @@
...
+		ARG 1 seed
+	METHOD m_lzlwtgyn nextGaussian ()D
+	METHOD m_mkdxraps nextBoolean ()Z
+	METHOD m_ngqtmirh createLegacy (J)Lnet/minecraft/unmapped/C_rlomrsco;
+		ARG 0 seed
+	METHOD m_paekwdby derive ()Lnet/minecraft/unmapped/C_rlomrsco;
+	METHOD m_schtxhxl range (II)I
+		ARG 1 min
+		ARG 2 max
+	METHOD m_sgbekski nextBetween (II)I
```
Zaeroses: I'm assuming this is a good mapping name, but just to ask anyone who feels like explaining: what's the difference between `range` and `nextBetween`?  These names, without any personal context with the code, mean to me that range will constrain random values to a certain range, while nextBetween would find the first random value that fits in a certain range.  What does the distinction between range and between signify?  Is it just to make it clear they are separate methods?

triphora (approved):

NoahvdAa (approved):

mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -0,0 +1,29 @@
...
+	METHOD m_djbqzhqs nextFloat ()F
+	METHOD m_ezprnxnd nextInt ()I
+	METHOD m_fakwfybt nextDouble ()D
+	METHOD m_ifdmmgky nextLong ()J
+	METHOD m_imwpekuz createUnsafe ()Lnet/minecraft/unmapped/C_rlomrsco;
+	METHOD m_iwfdczgh setSeed (J)V
+		ARG 1 seed
+	METHOD m_lzlwtgyn nextGaussian ()D
+	METHOD m_mkdxraps nextBoolean ()Z
+	METHOD m_ngqtmirh createLegacy (J)Lnet/minecraft/unmapped/C_rlomrsco;
```
Zaeroses, line 19: Is this [`createLegacy`] made to be consistent with other mappings elsewhere that are named the same? It doesn't seem clear enough to me as to what exactly it describes - I presume it is creating a legacy version of a random generator, but you might want to double-check and make sure it can't be confused between some different meanings:

- Creates an instance of RandomGenerator, with legacy options
- Creates an instance of a separate class, e.g. LegacyRandomGenerator (though I imagine that would be `createLegacyRandomGenerator`?)
- Creates a legacy/report of its current state
- Creates a 'legacy' entry in something that keeps track of the history of random generators

You can mostly rule these out logically, but I'm thinking that it has the possibility of being mistaken or confused, e.g. with beginners or just if someone's tired.  Do you think this is a valid concern, or does this meet other standards consistently enough for it not to be a worry?

Zaeroses (commented):
This is likely not a concern, but I thought it would be worth mentioning. Just a _potential_ ambiguity with the name, you might want to just comment on why this isn't an issue, or if any change is warranted.

mappings/net/minecraft/util/random/RandomGenerator.mapping
OroArmor: It's the second bullet. No need for the full name however.

mappings/net/minecraft/util/random/RandomGenerator.mapping
OroArmor: After checking the code the second should be `inclusiveRange`

mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -0,0 +1,29 @@
...
+		ARG 0 seed
+	METHOD m_paekwdby derive ()Lnet/minecraft/unmapped/C_rlomrsco;
+	METHOD m_schtxhxl range (II)I
+		ARG 1 min
+		ARG 2 max
+	METHOD m_sgbekski nextBetween (II)I
+		ARG 1 min
+		ARG 2 max
```
OroArmor:
```suggestion
	METHOD m_schtxhxl rangeClosed (II)I
		ARG 1 minInclusive
		ARG 2 maxInclusive
	METHOD m_sgbekski range (II)I
		ARG 1 minInclusive
		ARG 2 maxExclusive
```

PR #160, opened 2022-04-21
Switch inaccurate BuiltInStructureSets to MojMap's StructureSetKeys

TelepathicGrunt:
I think I did this pr correctly but lmk if it is wrong. Basically, this class holds only registry keys for structure sets. Not the actual structure sets (the class StructureSets holds the structure sets). This makes the original name no good. Solution, switch to MojMap name that is more clear that it is holding keys.

(change will need to be propagated to snapshot versions too. Do I need to make a pr for snapshot too?)

### DIFF
--- a/mappings/net/minecraft/world/gen/structure/StructureSetKeys.mapping
+++ b/mappings/net/minecraft/world/gen/structure/StructureSetKeys.mapping
@@ -1,3 +1,3 @@
-CLASS net/minecraft/unmapped/C_vreaftov net/minecraft/world/gen/structure/BuiltInStructureSets
+CLASS net/minecraft/unmapped/C_vreaftov net/minecraft/world/gen/structure/StructureSetKeys
 	METHOD m_nehwwprb register (Ljava/lang/String;)Lnet/minecraft/unmapped/C_xhhleach;
 		ARG 0 id


triphora (approved):

OroArmor:
@ByMartrixx is it possible to include this in one of your structure prs?

OroArmor (approved):

LambdAurora (commented):
Any reason to drop the BuiltIn part?

TelepathicGrunt:
The BuiltIn prefix never did make any sense to me. Anything in code for vanilla stuff is already "built in" so it is redundant and conveys no new info. If a modder wants to find the keys for vanilla stuff, they would try typing \<Thing\>Keys so in this case, if they want the key for vanilla Structure Sets, they would try StructureSetKeys first and get it right away. 

It's the same reason we do not have BuiltInBlocks because it is already implied that everything in Blocks is vanilla's stuff. And not using "BuiltIn" also matches the other key classes in quilt mapping such as `BiomeKeys` and `NoiseParametersKeys`

LambdAurora:
> The BuiltIn prefix never did make any sense to me. Anything in code for vanilla stuff is already "built in" so it is redundant and conveys no new info. If a modder wants to find the keys for vanilla stuff, they would try typing <Thing>Keys so in this case, if they want the key for vanilla Structure Sets, they would try StructureSetKeys first and get it right away.
> 
> It's the same reason we do not have BuiltInBlocks because it is already implied that everything in Blocks is vanilla's stuff. And not using "BuiltIn" also matches the other key classes in quilt mapping such as `BiomeKeys` and `NoiseParametersKeys`

So far builtin stuff is mostly used for stuff that can be modified through datapacks.

JaaiDead (approved):

triphora:
Why was this `update-base`'d? This was supposed to target 1.18.2 release

OroArmor:
oops

NoComment1105 (approved):

PR #194, opened 2022-06-02
Client option mappings

IotaBread:
Fixes #190 and #193

PR #197, opened 2022-06-05
Refactor Text methods and add more documententation.

LambdAurora:
While there's some overlap with #194, the new proposed names are more similar to what mojmap and yarn are doing.

The `create` prefix is superfluous in this context.
We also can compare this with brigadier, where methods are named similarly and not with a `create` prefix.
It also makes method chaining much nicer:
`Text.literal("Hello ").append(Text.translatable("world.translation").formatted(Formatting.GOLD)).append("!")`

Also rename copy and shallowCopy because those names were always confusing to modders and a large number of mistakes have been done because of those.

### DIFF
--- a/mappings/net/minecraft/text/Text.mapping
+++ b/mappings/net/minecraft/text/Text.mapping
@@ -1,56 +1,103 @@
 CLASS net/minecraft/unmapped/C_rdaqiwdt net/minecraft/text/Text
 	COMMENT A text. Can be converted to and from JSON format.
-	COMMENT
-	COMMENT <p>Each text has a tree structure, embodying all its {@link
+	COMMENT <p>
+	COMMENT Each text has a tree structure, embodying all its {@link
 	COMMENT #getSiblings() siblings}. To iterate contents in the text and all
 	COMMENT its siblings, call {@code visit} methods.
-	COMMENT
-	COMMENT <p>This interface does not expose mutation operations. For mutation,
+	COMMENT <p>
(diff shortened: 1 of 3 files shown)


triphora (approved):

OroArmor (approved):

NoComment1105 (approved):

IotaBread (approved):

Jamalam360 (approved):
Looks nice

mappings/net/minecraft/text/Text.mapping
```diff
@@ -1,56 +1,105 @@
...
+	METHOD m_kuqyhsxe selector (Ljava/lang/String;Ljava/util/Optional;)Lnet/minecraft/unmapped/C_npqneive;
+		COMMENT Creates a new mutable text of a resolveable entity selector.
+		COMMENT
+		COMMENT @return the new mutable selector text
+		ARG 0 selector
+			COMMENT the selector
 		ARG 1 separator
-	METHOD m_nxgujqbg createKeybind (Ljava/lang/String;)Lnet/minecraft/unmapped/C_npqneive;
-		ARG 0 text
+			COMMENT the optional seperator if there's multiple match issued from the selector
```
Leo40Git:
```suggestion
			COMMENT the optional seperator if there's multiple matches issued from the selector
```

Leo40Git (approved):
LGTM, except for this one tiny thing...

PR #202, opened 2022-06-29
Refactor and make boolean return wording more consistent.

LambdAurora:
One day we got a talk about how boolean returns/param javadocs should be worded.

We talked and talked, we threw a `else {@code false}` but no that's not English! We threw `otherwise {@code false}` but that's still wrong!

We ended up with the `or {@code false} otherwise` wording, it's very explicit, and more exact in the context of the English language.

And here we are, finally applying this to known documentation that used the "old" wording.

I'm honestly surprised by the few changes, tbh I went with a replace function so maybe I missed cases that I wasn't aware of.


OroArmor (approved):

IotaBread (approved):
Some of these could be changed to use `{@return}` tags, but I guess that could be left to another pr

triphora (approved):

NoComment1105 (approved):

PR #227, opened 2022-08-14
Add Intermediary Publication

### LINKED ISSUE #224: Publish Intermediary Quilt Mappings.
This would eliminate the need of Quilt Mappings on Loom which has been proven unstable.

IotaBread:
Publishes a version of QM with intermediary instead of hashed. Closes #224.

Still needs testing


LambdAurora:
Would it be possible to make building with Intermediary optional?
So if QM ever updates before Intermediary we can still publish a hashed version, and also maybe have a way to re-publish the mappings when an Intermediary artifact is missing.

IotaBread:
I'm gonna try to figure out if it's possible with a gradle dependency (as it is currently), otherwise it'll have to be a hardcoded url

LambdAurora:
> I'm gonna try to figure out if it's possible with a gradle dependency (as it is currently), otherwise it'll have to be a hardcoded url

The other option is to figure out something with GitHub actions ahah.
Not sure how tho, but we definitely will need an additional action to republish to intermediary.

IotaBread:
Yeah, we would need an action to republish only intermediary, which I'm afraid might cause some trouble with build numbers. About using gradle or hardcoding, what I meant is that I'm not sure gradle would like it when intermediary isn't in any maven repo

OroArmor (approved):
Looks good so far

triphora (approved):
DEATH TO QMOL!

I'll be able to test this out in a week's time, assuming it hasn't already been merged by then.

OroArmor (approved):

LambdAurora (approved):

PR #236, opened 2022-08-21
change `hasTag` in `Holder` to `isIn` for greater consistency

eristhea:
this pr removes the `hasTag` name and changes it to `isIn`, making all methods that check tags have consistent naming.

(original pr comment) `isIn` is a much worse mapping than `hasTag`, which is both already used in quilt mappings and conveys more information (shows the user that it's checking in a tag without making them look at parameters).

### DIFF
--- a/mappings/net/minecraft/block/AbstractBlock.mapping
+++ b/mappings/net/minecraft/block/AbstractBlock.mapping
@@ -366,7 +366,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 			ARG 2 pos
 			ARG 3 flags
 			ARG 4 maxUpdateDepth
-		METHOD m_lyoypzum hasTagAndMatches (Lnet/minecraft/unmapped/C_ednuhnnn;Ljava/util/function/Predicate;)Z
+		METHOD m_lyoypzum isInAndMatches (Lnet/minecraft/unmapped/C_ednuhnnn;Ljava/util/function/Predicate;)Z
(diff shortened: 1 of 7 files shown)


supersaiyansubtlety (commented):
I'd prefer `isInTag` if you don't want to rely on params for understanding.

`hasTag` implies a backwards relationship. Tags have/contain entries, the entries don't "have" the tag.

LambdAurora (commented):
imo `isIn` is perfectly fine, when you read a code it translates well what's happening.

I have a block, and a stone tag
`block.isIn(Block tags.STONES)` translates naturally and efficiently the intent.

Oh and I would consider the use of the verb `has` as entirely wrong: the idea of a tag is it lists game objects, while sure the game objects will contain a list of tags, it's a list of tags it's in, *not the tags it has*. The tag *has* a game object, and the object *is in* a tag.

NoComment1105 (approved):

eristhea:
The original intent of this PR was more for consistency and less for improving names. Since it seems `isIn` is the preferred name, I'll rework this PR to change `hasTag` to `isIn`. I just saw a consistency issue and picked my preferred name.

edit: when picking which name to change I did consider the slightly misleading nature of saying "has", but I thought it was fine simply because things like "this block has three tags" is commonly said.

eristhea:
Changes made!
This is a one line pr now but it's still worth the fix to me.

LambdAurora (approved):

supersaiyansubtlety (approved):

PR #239, opened 2022-08-27
map `TreeDecorator`'s inner class

eristhea:
fairly straightforward, but I'm not sure about the name of the entire class. I was considering `Generator`, but since it's replacing already existing blocks instead of generating new ones, I thought `BlockReplacer` was a good name. Please leave feedback on that specifically though!

### DIFF
--- a/mappings/net/minecraft/world/gen/treedecorator/TreeDecorator.mapping
+++ b/mappings/net/minecraft/world/gen/treedecorator/TreeDecorator.mapping
@@ -2,4 +2,24 @@ CLASS net/minecraft/unmapped/C_wyluesrv net/minecraft/world/gen/treedecorator/Tr
 	COMMENT Tree decorators can add additional blocks to trees, such as vines or beehives.
 	FIELD f_ytzktbtf TYPE_CODEC Lcom/mojang/serialization/Codec;
 	METHOD m_gnmhbdck generate (Lnet/minecraft/unmapped/C_wyluesrv$C_jvnizkzw;)V
+		ARG 1 context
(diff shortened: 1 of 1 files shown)


OroArmor (approved):

IotaBread (approved):

triphora (approved):

Kroppeb:
I personally don't like the name "BlockReplacer" at all. When it's passed as argument you call it "context" which seems to agree with my opinion.

supersaiyansubtlety (approved):

eristhea:
the more I think about it the more I dislike the name
but since there's data being sent to it, I don't agree with `Context` either
@Kroppeb how would you feel about simplifying it to `Placer`?

eristhea:
I added some javadoc and decided to go with the `Placer` name instead. I think it's a better name since this class is mostly replacing air, and the `Block` prefix felt redundant, given that the user is providing it BlockStates.

Kroppeb:
mmh, The thing that kinda annoys me is the fact that the class doesn't really do much appart from being a wrapper around data. The only actual logic is the fact it sorts the lists and provides a utility method for placing vines. In the mean time, `Placer` doesn't convey the fact it is also used to store the roots, leaves and logs locations.

eristhea:
exactly why I haven't been able to decide on a name, it's just such an odd class.
do you have any suggestions?

NoComment1105 (approved):

NoahvdAa (approved):
Looks good from web

AlsoSylv (approved):

PR #245, opened 2022-10-10
Rename some freezing related mappings

Platymemo:
`slowMovement` -> `setMovementMultiplier`: `slowMovement` is a bit of a misnomer because it directly applies a movement multiplier. It could increase entity speed too. It just happens to be used by Mojang purely for slowing (ish, powder snow provides a 1.5x vertical multiplier on powder snow)

`onLanding` -> `resetFallDistance`: `onLanding` is slightly inaccurate, because it is called by the above method and not just when actually landing on a block. Its primary purpose is just to reset the fall distance (it also sets the stat for a player's highest fall distance if their pre-reset fall distance is greater than their best)

`getMinFreezeDamageTicks` -> `getMaxFreezeTicks`: It is the threshold for where freeze starts damaging yes, but it's also the maximum value that freeze ticks can be set to.


IotaBread (approved):

LambdAurora (approved):

OroArmor (approved):

OroArmor:
I think this can be merged

Issue #251, opened 2022-10-19
`RegistryWorldView` no longer exposes access to a registry manager


IotaBread:
As such, its current name is wrong and should be changed. Its mojmap name is `CommonLevelAccessor`, and exposes the following methods:
- `getBlockEntity` from `WorldView`
- `getEntityCollisions` from `EntityView`
- `doesNotIntersectEntities` from `EntityView`
- `getTopPosition` from `WorldView`

eristhea:
a couple possible names:
- `EntityWorldView`, since it exposes methods to get entities. Questionable name, since only half the methods are used to get entities, and one is completely unrelated.
- `(Basic|Base|Simple)WorldView`: a decent name (at the very least, better than the current misleading one), but doesn't communicate much.

PR #252, opened 2022-10-19
Fix javadocs for 22w42a

IotaBread:
Also map stuff in `ChatLog`. As for #251, I just removed the javadoc, it still needs a new name

### DIFF
--- a/mappings/net/minecraft/block/Block.mapping
+++ b/mappings/net/minecraft/block/Block.mapping
@@ -103,8 +103,6 @@ CLASS net/minecraft/unmapped/C_mmxmpdoq net/minecraft/block/Block
 		ARG 2 stack
 	METHOD m_fcwpdfta appendStacks (Lnet/minecraft/unmapped/C_lnokcayd;Lnet/minecraft/unmapped/C_rnrfftze;)V
 		COMMENT Appends the stacks of this block shown in the item group to the list.
-		COMMENT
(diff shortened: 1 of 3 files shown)


OroArmor (approved):

mappings/net/minecraft/block/Block.mapping
```diff
@@ -103,8 +103,6 @@ CLASS net/minecraft/unmapped/C_mmxmpdoq net/minecraft/block/Block
 		ARG 2 stack
 	METHOD m_fcwpdfta appendStacks (Lnet/minecraft/unmapped/C_lnokcayd;Lnet/minecraft/unmapped/C_rnrfftze;)V
 		COMMENT Appends the stacks of this block shown in the item group to the list.
-		COMMENT
-		COMMENT @see net.minecraft.item.BlockItem#appendStacks(ItemGroup, DefaultedList)
```
LambdAurora, line 107: Is there no alternative?

LambdAurora (approved):
Aside from my comment, all good.

mappings/net/minecraft/block/Block.mapping
OroArmor: i think we should wait until the new item group stuff is mapped to know for sure

IotaBread: I didn't really check, but because of the item-group changes an alternative may not exist or may be unmapped, so as Oro said, we should wait until the new stuff is mapped

PR #268, opened 2022-10-31
Rename ScoreHolderArgumentType methods and add JavaDoc

### LINKED ISSUE #205: Rename ScoreHolderArgumentType.getScoreboardScoreHolders
The difference between getScoreHolders and this is that it returns all players if given a `*` wildcard instead of throwing an exception.

Mojmap's getNamesWithDefaultWildcard is closer to the truth, though still not ideal: "default" is actually "all known players" here.

Cypher121:
Closes #205. Added JavaDoc to clear up some earlier confusion about the exact behavior of each variant of `getScoreHolders`.


OroArmor (approved):

NoComment1105 (changes requested):
Actions fail for this due to the `@param` annotations. `@param` annotations need to be plan annotations on the param itself. Engima has an option for this, just right click the param, doc the doc and then it will sort itself out.

NoComment1105 (approved):

PR #267, opened 2022-10-30
Random mappings (supposedly targetting items and blocks)

EnnuiL:
I've tried my best to map a lot of the unmapped stuff on both 1.19.3 and earlier; please review this thoroughly

PR #269, opened 2022-11-03
Map several registry-related classes.

LambdAurora:


mappings/net/minecraft/util/registry/SimpleRegistry.mapping
```diff
@@ -7,7 +7,7 @@ CLASS net/minecraft/unmapped/C_secmvxxe net/minecraft/util/registry/SimpleRegist
 	FIELD f_oyicsmaj byId Ljava/util/Map;
 	FIELD f_rbivtjbc nextId I
 	FIELD f_tjecakfq lifecycle Lcom/mojang/serialization/Lifecycle;
-	FIELD f_tlngcayg entryToLifecycle Ljava/util/Map;
+	FIELD f_tlngcayg getLifecycle Ljava/util/Map;
```
LambdAurora:
```suggestion
	FIELD f_tlngcayg entryToLifecycle Ljava/util/Map;
```

Damn you enigma.

mappings/net/minecraft/data/server/tag/BlockTagsProvider.mapping
```diff
@@ -0,0 +1 @@
+CLASS net/minecraft/unmapped/C_wayypnpx net/minecraft/data/server/tag/BlockTagsProvider
```
EnnuiL, line 1: Oh, I've touched those providers on #267; I feel like using `BaseBlocksTagProvider` and `FutureBlockTagsProvider` might be better than `BlocksTagProvider` and `UpdateOneTwentyBlockTagsProvider`; i really don't like the mojmap name for the latter (although to be fair, it's only an opinion that might be bad if they add concurrent ones) and you'll might want consistency with the item providers, which had the same split except that they both share a common source that isn't abstract either

EnnuiL (changes requested):
The PR seems good, even though I'm not really a mappings person; I did notice, however, some spots in which I'm better equipped to talk about

OroArmor (approved):

NoComment1105 (approved):

mappings/net/minecraft/data/server/tag/BlockTagsProvider.mapping
LambdAurora: I'm not entirely sure about the FutureBlockTagsProvider

Even if then in 1.20 there's one for 1.21 stuff, the mapping will be lost due to mojmap having very different names.
I'm also wondering what would happens if there's ever 2 concurrent experiments (which kinda already happened with the combat updates, etc)?

EnnuiL: yeah, after thinking about it, `FutureBlockTagsProvider` wouldn't really be great; The rest of the feedback might still be worth investigating though, but yeah, maybe `UpdateOneTwenty` is the better name for now (unless Mojang ends revealing that update name :P)

PR #276, opened 2022-11-20
consolidate registry package and improve registry/tag mappings

eristhea:
this pr does four things:
1. implements yarn's [great registry repackaging](https://gist.github.com/apple502j/1bc2dd67fe7e055d9be5d00fa0c04fbe#great-registry-repackaging) to keep all registry packages in one place
2. names all tags. Many tags have been named differently from their filenames, which could cause confusion, but increases clarity in code. Please provide feedback on this change!
3. names all registries. Most registries match what they used to be in 1.19.2.
4. does miscellaneous new mappings, all confined to the registry package


OroArmor (approved):

eristhea:
this may still get some big changes; I'm planning on reading through all the files again tomorrow to make sure all the names are good

LambdAurora:
What's the reasoning for having the tag package in the registry package?

eristhea:
The same reasoning as yarn had. Tags are all registry-based, so putting them in the registry package is a good reflection of how they work internally.

NoComment1105 (approved):
Far better than what I produced 😅

eristhea:
second check over complete; I'm comfortable with putting the ix0rai seal of approval :tm: on this!

PR #279, opened 2022-11-20
entity variant -> entity type

eristhea:
today on things no one but me cares about

notes:
in the registry, variant -> type refactor is done in #276 
cats, frogs, and paintings are affected
this pr updates the conventions, adding
> Entities will occasionally have different appearances based on their properties - for example, cats can have different fur colours.

> When dealing with things like this, use `Variant` for cosmetic-only distinctions while using `Type` is used when there is behavior associated with the appearance change. A general rule for this is to check if the different types are stored in a registry. If they are, you should be using `Type` over `Variant`.

### DIFF
--- a/mappings/net/minecraft/entity/passive/CatEntity.mapping
+++ b/mappings/net/minecraft/entity/passive/CatEntity.mapping
@@ -1,5 +1,5 @@
 CLASS net/minecraft/unmapped/C_rxtmregr net/minecraft/entity/passive/CatEntity
-	COMMENT Meow.
+	COMMENT mrrp nya
 	FIELD f_crlwzfoc HEAD_DOWN Lnet/minecraft/unmapped/C_rinmcaxy;
 	FIELD f_czqfqakz prevSleepAnimation F
 	FIELD f_ebcwdxuh tailCurlAnimation F
@@ -7,7 +7,7 @@ CLASS net/minecraft/unmapped/C_rxtmregr net/minecraft/entity/passive/CatEntity
 	FIELD f_gaahrhec TEMPTED_SPEED D
 	FIELD f_hqvkpijz fleeGoal Lnet/minecraft/unmapped/C_rxtmregr$C_jlusltvy;
 	FIELD f_iywwshas sleepAnimation F
-	FIELD f_jycwwapx VARIANT Lnet/minecraft/unmapped/C_rinmcaxy;
+	FIELD f_jycwwapx TYPE Lnet/minecraft/unmapped/C_rinmcaxy;
 	FIELD f_krqstxdp headDownAnimation F
 	FIELD f_lfmjmddd COLLAR_COLOR Lnet/minecraft/unmapped/C_rinmcaxy;
 	FIELD f_lsvomhnh SPRINTING_SPEED D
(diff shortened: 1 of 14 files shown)


OroArmor (approved):

IotaBread (approved):

eristhea:
@EnnuiL mentioned in discord that there are actually more usages of variant that I missed in my initial enigma search, so I'm converting this to draft until I've done a more thorough search.

eristhea:
point to consider: `FrogEntity` uses `getFrogType`, which is a convention I followed instead of using `getType` everywhere. However, this led to things like `getTropicalFishType`, which is significantly longer than simply `getVariant`. Should we use `getType`, or `getEntityNameType`? I'm leaning towards the first.

OroArmor:
I think variant for something that has a registry unless the registry types have behavior associated with them, and type for something that is just an enum.  I dont want to change `BlockEntityType` and `EntityType`, which is why i think the clause for registries is needed. If this could be added to the conventions file in this pr that would be great.

OroArmor:
Also frogs should 100% be using variant. i just checked the code, and i messed up on that mapping

eristhea:
hm.
I'm thinking we use `variant` for something that's purely cosmetic, and `type` for anything with behaviour associated to it. As far as I know without checking, that would discard all this PR's content except for paintings, and would require frogs being refactored from type *to* variant.

OroArmor:
I think so as well, but i also believe it makes better mappings and is more intuitive to new modders.

eristhea:
that way we have a simple explanation for `BlockEntityType` and `EntityType` keep their behaviour, as well as a very simple convention for deciding between the two. Since things like frogs and cats do have a registry, I think we should avoid the registry clause.

OroArmor:
maybe something like:
> Use `Variant` for cosmetic-only distinctions while `Type` is used when there is significant behavior associated with it.

eristhea:
looking into things further:
I was wrong about the registry bit not being needed.
The four entities that have variant registries, villagers, cats, frogs, and paintings, all have slightly different behaviour based on their variant. Frogs and cats have different spawning, villagers have their professions, and paintings change size. 

With that in mind:
> Use `Variant` for cosmetic-only distinctions while `Type` is used when there is ~~significant~~ any behavior associated with it. A general rule for this is to check if the different types are stored in a registry. If they are, you should be using `Type` over `Variant`.

eristhea:
setting this back to draft while we discuss

eristhea:
@OroArmor thoughts on that snippet as the convention?

OroArmor:
Sounds good. I'm curious what @ByMartrixx and @LambdAurora think.

eristhea:
perfect!
I'm going to make the changes, and keep the pr as draft until their opinions are in.

mappings/net/minecraft/entity/passive/CatEntity.mapping
```diff
@@ -1,13 +1,13 @@
 CLASS net/minecraft/unmapped/C_rxtmregr net/minecraft/entity/passive/CatEntity
-	COMMENT Meow.
+	COMMENT mrrp nya
```
NoComment1105: lmao

NoComment1105 (approved):
LGTM 👍🏼

mappings/net/minecraft/entity/passive/CatEntity.mapping
eristhea: nooooo I was trying to slip it in unnoticed

eristhea:
mojang reworked entity variants; closing in favour of a new pull request

PR #284, opened 2022-11-22
map VariantProvider and clean up variant <-> type distinction

eristhea:
the sequel to the sadly deceased #279 
this pr gives up on having a convention and simply accepts mojang's `variant` name, since the new interface-based structure makes it impossible to have a clear distinction between variants and types


OroArmor (approved):

OroArmor:
Should we also consider renaming the enums?

eristhea:
of course; they slipped my mind in my haze of grief over my previous pr

OroArmor (changes requested):
I would look for uses of BoatVariant and make sure they are changed from type

eristhea:
good idea!
will take a look

eristhea:
will be more thorough tomorrow; but I changed all the obvious ones

supersaiyansubtlety (approved):

eristhea:
oh, note from tomorrow:
couldn't find any more; I did make a check

OroArmor (approved):

PR #285, opened 2022-11-23
improve world-related mappings

eristhea:
wip pr to increase coverage around world things

just opening so that what I'm working on is visible (sorry nocomment, I stole `NeighborUpdater` from you)

### DIFF
--- a/mappings/net/minecraft/world/entity/EntityLookup.mapping
+++ b/mappings/net/minecraft/world/entity/EntityLookup.mapping
@@ -4,10 +4,12 @@ CLASS net/minecraft/unmapped/C_mmgrdxia net/minecraft/world/entity/EntityLookup
 	COMMENT <p>It supports iteration, unfiltered or via intersection with boxes and/or type filtration, and accessing by ID.
 	COMMENT
 	COMMENT @param <T> the type of indexed entity
-	METHOD m_hcbcytkb forEachIntersects (Lnet/minecraft/unmapped/C_hbcjzgoe;Ljava/util/function/Consumer;)V
+	METHOD m_hcbcytkb forEachIntersecting (Lnet/minecraft/unmapped/C_hbcjzgoe;Ljava/util/function/Consumer;)V
+		COMMENT Runs the supplied consumer over every cached entity that intersects with the provided box.
 		ARG 1 box
 		ARG 2 action
-	METHOD m_jinuyjwx forEachIntersects (Lnet/minecraft/unmapped/C_ocpldaah;Lnet/minecraft/unmapped/C_hbcjzgoe;Lnet/minecraft/unmapped/C_gibrwunu;)V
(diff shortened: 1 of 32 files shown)


eristhea:
pretty happy with the amount of content mapped here; going to say this is ready

mappings/net/minecraft/world/entity/EntityLookup.mapping
```diff
@@ -1,17 +1,24 @@
...
 	COMMENT An interface for looking up entities.
 	COMMENT
-	COMMENT <p>It supports iteration, accessing by ID, or by intersection with boxes.
+	COMMENT <p>It supports iteration, unfiltered or via intersection with boxes and/or type filtration, and accessing by ID.
 	COMMENT
 	COMMENT @param <T> the type of indexed entity
 	METHOD m_hcbcytkb forEachIntersects (Lnet/minecraft/unmapped/C_hbcjzgoe;Ljava/util/function/Consumer;)V
 		ARG 1 box
 		ARG 2 action
+	METHOD m_jinuyjwx forEachIntersects (Lnet/minecraft/unmapped/C_ocpldaah;Lnet/minecraft/unmapped/C_hbcjzgoe;Lnet/minecraft/unmapped/C_gibrwunu;)V
```
supersaiyansubtlety:
```suggestion
	METHOD m_jinuyjwx forEachIntersect (Lnet/minecraft/unmapped/C_ocpldaah;Lnet/minecraft/unmapped/C_hbcjzgoe;Lnet/minecraft/unmapped/C_gibrwunu;)V
```
or
```suggestion
	METHOD m_jinuyjwx forEachIntersected (Lnet/minecraft/unmapped/C_ocpldaah;Lnet/minecraft/unmapped/C_hbcjzgoe;Lnet/minecraft/unmapped/C_gibrwunu;)V
```
?

supersaiyansubtlety (changes requested):

mappings/net/minecraft/world/entity/EntityLookup.mapping
eristhea: hmm, why?
I think the `forEachIntersects` name is better than both of those
`forEachIntersect` sounds like it's going to be making something intersect, and `forEachIntersected` is past tense, which is odd since the entity and box are currently intersecting

supersaiyansubtlety: IIUC, "intersects" is a verb here and the full sentence the method name and params imply would be  "For each [entity the box] intersects[, run the consumer.]"

However "intersects" in this context *sounds* like a plural noun, which would be grammatically incorrect: "For each intersects [between the the box and entities, run the consumer.]" (which is what I initially thought when I read it)

If I'm correct and it's a verb, then my suggestion of "intersect" was wrong.

"intersected" avoids the confusion with a plural noun, but it is past tense for no good reason.

Maybe "intersecting" would be better?

eristhea: > IIUC, "intersects" is a verb here and the full sentence the method name and params imply would be "For each [entity the box] intersects[, run the consumer.]"

yes, that's exactly what it is
I like `intersecting`, I'll go with that
thank you!

supersaiyansubtlety (approved):

OroArmor (approved):

PR #289, opened 2022-11-28
Updates dependencies

OroArmor:
Also makes it so we know what unpick values fail, only took 2 years :iea:


buildSrc/src/main/java/quilt/internal/tasks/unpick/UnpickJarTask.java
```diff
@@ -24,7 +24,6 @@ public UnpickJarTask() {
 
         this.getMainClass().set("daomephsta.unpick.cli.Main");
         classpath(getProject().getConfigurations().getByName("unpick"));
-        systemProperty("java.util.logging.config.file", getProject().file("unpick-logging.properties"));
```
EnnuiL, line 27: The file can be safely removed now, we should do that

EnnuiL (changes requested):
Just one little thing

EnnuiL (approved):

supersaiyansubtlety (approved):

IotaBread (approved):

eristhea (approved):

PR #294, opened 2022-12-01
Random Note Block-related mappings

EnnuiL:
Few notes about this PR:
 - There were two Instrument classes; Let's cut it down to 1
 - I chose animation ticks here because it wasn't clamped to 1.0 but then I saw that there isn't really a convention for this; One should be established, with animation progress representing a 0% to 100% (0.0 to 1.0) progress and animation ticks being able to go above and beyond
 - `setHeadRotation` is definitely wrong, but the convention for setting animation progress and rotations are, uh, `setAngles`; this should be fixed
 - I was going to do the Holder-related mappings but I ended deciding it would be too hard for me


supersaiyansubtlety (approved):

mappings/net/minecraft/block/NoteBlock.mapping
```diff
@@ -2,9 +2,11 @@ CLASS net/minecraft/unmapped/C_nzuwmjdt net/minecraft/block/NoteBlock
 	FIELD f_ajsbzucp NOTE Lnet/minecraft/unmapped/C_vltzvhxi;
 	FIELD f_aucyqdcc INSTRUMENT Lnet/minecraft/unmapped/C_cgckxfsw;
 	FIELD f_kmjcamoe POWERED Lnet/minecraft/unmapped/C_xhwijdsd;
+	FIELD f_zcrvmfco NOTE_VOLUME I
 	METHOD m_eelivhoc getStateWithInstrument (Lnet/minecraft/unmapped/C_vdvbsyle;Lnet/minecraft/unmapped/C_hynzadkk;Lnet/minecraft/unmapped/C_txtbiemp;)Lnet/minecraft/unmapped/C_txtbiemp;
 		ARG 1 worldAccess
-	METHOD m_umbeungg mobHeadSoundsEnabled (Lnet/minecraft/unmapped/C_vdvbsyle;)Z
+	METHOD m_iwpuznep getPlayerHeadNoteSound (Lnet/minecraft/unmapped/C_cdctfzbn;Lnet/minecraft/unmapped/C_hynzadkk;)Lnet/minecraft/unmapped/C_ncpywfca;
```
eristhea: could be made more concise, I think the `Note` is fairly redundant here

eristhea (approved):

mappings/net/minecraft/block/NoteBlock.mapping
EnnuiL: Ah, I was worried that it would be too concise to the point where it loses a bit of meaning; Maybe I could either remove `Note` or `Sound`...? maybe the former

eristhea: personal preference would be note, but either removal works!

supersaiyansubtlety (approved):

mappings/net/minecraft/client/render/block/entity/model/DragonHeadEntityModel.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/unmapped/C_ksfxmjly net/minecraft/client/render/entity/model/DragonHeadEntityModel
+CLASS net/minecraft/unmapped/C_ksfxmjly net/minecraft/client/render/block/entity/model/DragonHeadEntityModel
```
OroArmor: This should be in `net/minecraft/client/render/entity/model/` like `net/minecraft/client/render/entity/model/SkullEntityModel`

mappings/net/minecraft/client/render/block/entity/model/PiglinHeadEntityModel.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/unmapped/C_eebmnhjg net/minecraft/client/render/block/entity/PiglinHeadEntityModel
+CLASS net/minecraft/unmapped/C_eebmnhjg net/minecraft/client/render/block/entity/model/PiglinHeadEntityModel
```
OroArmor: Same as the other head model

mappings/net/minecraft/client/render/block/entity/model/SkullBlockEntityModel.mapping
```diff
@@ -1,5 +1,5 @@
-CLASS net/minecraft/unmapped/C_tuhbiezx net/minecraft/client/render/block/entity/SkullBlockEntityModel
-	METHOD m_tzqjhnwe setHeadRotation (FFF)V
+CLASS net/minecraft/unmapped/C_tuhbiezx net/minecraft/client/render/block/entity/model/SkullBlockEntityModel
```
OroArmor: Same as the other head model

OroArmor (changes requested):
Only a small change.

mappings/net/minecraft/client/render/block/entity/model/DragonHeadEntityModel.mapping
EnnuiL: It's a model that is only used on BlockEntities though, so having them on the entity directory doesn't really make sense

EnnuiL: I found out that my restructuration was flawed, so I did it again but better

supersaiyansubtlety (approved):

OroArmor (approved):

PR #300, opened 2022-12-04
rename registry classes for 1.19.3

eristhea:
fairly urgent pr to rename registry classes before 1.19.3

reasoning:
the current names don't make sense and don't communicate enough information.
`Registries`: stores registry keys. you'd expect it to store registries, wouldn't you? I've renamed to `RegistryKeys`.
`BuiltInRegistries`: overcomplicated name for something that simply stores registries. I've renamed this one to `Registries`.

notes:
large git diff is simply because the name of `Registries` is identical to the old name for `RegistryKeys`.


IotaBread (approved):

OroArmor (approved):

LambdAurora (approved):
tbh I wonder how those name came to be when there was old names similar to this model

PR #301, opened 2022-12-06
add mapping lint spellchecker

### LINKED ISSUE #47: Add spell checker
Builds off of #46 and FabricMC/Fabric-Filament#8

eristhea:
the spellchecker identifies all words used in mappings and compares them against a list of whitelisted words

features:
checks method names, class names, field names, javadocs, anything that's a mapping entry
has a whitelist that you can manually add to
ends britishisms as well as spelling mistakes

notes:
fixes #47 

todo:
- [x] caching for word list - waiting on @OroArmor 
- [x] organise minecraft word file
- [x] documentation


buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+public class SpellingChecker implements Checker<Entry<?>> {
+    private static final String PACKAGE_INFO_CLASS_PACKAGE = "net/minecraft/unused/packageinfo/";
+
+    private static final Set<String> ALLOWED_WORDS = new HashSet<>();
+    static {
+        // collect allowed words
+        URL url;
+        try {
+            url = new URL("https://raw.githubusercontent.com/ix0rai/qm-base-allowed-wordlist/main/allowed_english_words.txt");
+            Set<String> englishWords = new java.util.Scanner(url.openStream()).useDelimiter("\\A").next().lines().collect(Collectors.toSet());
```
IotaBread: This should use the getLines method

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+            ALLOWED_WORDS.addAll(englishWords);
+        } catch (Exception e) {
+            throw new RuntimeException("failed to download and read english word file for spell check in mapping lint task!");
```
IotaBread: The caught exception should be passed to the `RuntimeException`

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        Set<String> minecraftWords = getLines(parsedFile);
+        // ignore comments and empty lines
+        Set<String> minecraftWordsCopy = Set.copyOf(minecraftWords);
+        for (String word : minecraftWordsCopy) {
+            if (word.isEmpty() || word.startsWith("//")) {
+                minecraftWords.remove(word);
+            }
+        }
```
IotaBread: Not needed if you implement what I commented below

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+
+        ALLOWED_WORDS.addAll(minecraftWords);
+    }
+
+    private static Set<String> getLines(InputStream stream) {
+        return new java.util.Scanner(stream).useDelimiter("\\A").next().lines().collect(Collectors.toSet());
```
IotaBread: You could use `new BufferedReader(new InputStreamReader(stream)).lines().collect(Collectors.toSet())` instead of a Scanner. You can also filter out comment/empty lines with `.filter(s -> !s.isBlank() && !s.startsWith("//"))` before `.collect`

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+
+            // fixme optimise!
+            // split by slashes
+            String[] slashSeparatedWords = name.split("/");
+            for (String word : slashSeparatedWords) {
+                // split by underscores
+                String[] underscoreSeparatedWords = word.split("_");
+                for (String underscoreSeparatedWord : underscoreSeparatedWords) {
+                    // split by uppercase characters and preserve
+                    words.addAll(List.of(Arrays.stream(underscoreSeparatedWord.split("(?<!(^|[A-Z]))(?=[A-Z])|(?<!^)(?=[A-Z][a-z])")).map(String::toLowerCase).toArray(String[]::new)));
```
IotaBread:
```suggestion
                    words.addAll(Arrays.stream(underscoreSeparatedWord.split("(?<!(^|[A-Z]))(?=[A-Z])|(?<!^)(?=[A-Z][a-z])")).map(String::toLowerCase).toList());
```

I'd put the `word.split(<regex>)` in another variable. The regex here could also split on digits so you don't have to handle them later

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,109 @@
...
+        if (name != null
+                && !name.startsWith(PACKAGE_INFO_CLASS_PACKAGE)
+                // ignore unmapped methods and classes
+                && !(entry instanceof MethodEntry && name.startsWith("m_"))
+                && !(entry instanceof ClassEntry && name.startsWith("C_"))) {
+            Set<String> words = new HashSet<>();
+
+            // fixme optimise!
+            // split by slashes
+            String[] slashSeparatedWords = name.split("/");
```
IotaBread: Since only classes can have slashes in their target names, you could do something like this
```java
String[] packageNames = new String[0];
if (entry instanceof ClassEntry) {
  name = name.substring(name.lastIndexOf('/') + 1);
  packageNames = name.substring(0, name.lastIndexOf('/'));
}

words.addAll(List.of(packageNames));
// ...

String[] underscoreSeparatedWords = name.split("_");
```

eristhea: packages can have underscores in the names, so the underscore filter is still required on them

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: will look into it!

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
IotaBread: Ah, right. What if you put them into a collection alongside the name, and then handle underscores on that collection?

eristhea: good idea; I'll implement that!
any optimisation is important

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: ah, now I remember why I handled digits differently
if we include them in this regex, it preserves them in the split strings, and we still have to filter them out later

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final Set<String> ALLOWED_WORDS = new HashSet<>();
+    static {
+        // collect allowed words
+        URL url;
+        try {
+            url = new URL("https://raw.githubusercontent.com/ix0rai/qm-base-allowed-wordlist/main/allowed_english_words.txt");
+            Set<String> englishWords = getLines(url.openStream());
+            ALLOWED_WORDS.addAll(englishWords);
+        } catch (Exception e) {
+            throw new RuntimeException("failed to download and read english word file for spell check in mapping lint task!\n" + e);
```
IotaBread:
```suggestion
            throw new RuntimeException("failed to download and read english word file for spell check in mapping lint task!", e);
```

OroArmor (approved):

CONTRIBUTING.md
```diff
@@ -31,6 +31,27 @@ or in our other community spaces. We're always happy to help if you need us!
     raised during that time, leave a comment on your PR requesting for it to be merged. A Mappings team member will
     take a final look over your PR, and, if everything looks good, merge it!
 
+## Guide: Fixing Failing Builds
+There are two main reasons the build may fail on your pull request:
```
IotaBread: Should add a note that the build may fail for other reasons not listed here, and in such case, the pr author may ask for help to fix the issue

CONTRIBUTING.md
```diff
@@ -31,6 +31,27 @@ or in our other community spaces. We're always happy to help if you need us!
...
+        The most common is that the word is a valid word in the context of Minecraft, but is not in the normal english dictionary, such as "prismarine".
+        In this case, you can add the word to [QM's Minecraft dictionary](buildSrc/src/main/resources/minecraft_specific_words.txt) under the corresponding category, which in this case would be `items`.  
+        Another possibility is that you have written the word in British English, which is invalid since QM uses American English exclusively.
+        Here, just change the word to its American English equivalent.  
+      
+    After performing any of these fixes, run `./gradlew mappingLint` to check that the error is fixed.
+2. ### Javadoc Failure
+   A javadoc failure will occur when you rename a class or method that a piece of javadoc points to.
+   It gives the error `error: reference not found * {@link TheOldName}`.
+   To fix this, you need to change the javadoc to point to the new name. Simple!
```
IotaBread, line 53:
```suggestion
   To fix this, you need to change the javadoc to point to the new name. Simple!
   You can open the javadoc editor in Enigma by right-clicking the entry and selecting the `Edit Javadoc` option, or by pressing `Ctrl + D` by default.
```

buildSrc/src/main/java/quilt/internal/tasks/lint/MappingLintTask.java
```diff
@@ -13,6 +24,7 @@
 import cuchaz.enigma.translation.representation.entry.ClassEntry;
 import cuchaz.enigma.translation.representation.entry.Entry;
 import cuchaz.enigma.translation.representation.entry.MethodEntry;
+import javax.inject.Inject;
```
IotaBread, line 27: Unused import

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    private static void checkJavadoc(String javadoc, ErrorReporter errorReporter) {
+        if (javadoc != null && !javadoc.isEmpty()) {
+            // in this case we just have to run the world's slowest regex, there's no clever optimisations based on type to be had
+            List<String> splitJavadoc = splitWords(List.of(javadoc), word -> true, "[0-9_\\[\\]\\-\\.\\{\\}\\(\\)@#/\":; ,<>\n=\\*\\|\\+%`§&!]");
+            for (String word : splitJavadoc) {
+                if (!word.isBlank()) {
+                    checkWord(word.toLowerCase(), errorReporter, "javadoc");
```
IotaBread: I think it would be good to keep the casing for the error message

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,190 @@
...
+
+            checkMapping(entry, name, errorReporter);
+            checkJavadoc(mapping.javadoc(), errorReporter);
+        }
+    }
+
+    private static void checkJavadoc(String javadoc, ErrorReporter errorReporter) {
+        if (javadoc != null && !javadoc.isEmpty()) {
+            // in this case we just have to run the world's slowest regex, there's no clever optimisations based on type to be had
+            List<String> splitJavadoc = splitWords(List.of(javadoc), word -> true, "[0-9_\\[\\]\\-\\.\\{\\}\\(\\)@#/\":; ,<>\n=\\*\\|\\+%`§&!]");
```
IotaBread: Why isn't this just `"[^A-Za-z]"`? The current regex also has a lot of redundant escapes, so in reality it looks like this ```"[0-9_\[\]\-.{}()@#/":; ,<>\n=*|+%`§&!]"```

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param strippingRegex a regex to split words by
+     * @return a list of split words
+     */
+    private static List<String> splitWords(List<String> words, Predicate<String> strippingCheck, String strippingRegex) {
+        // split by uppercase characters and numbers and preserve
+        List<String> splitNames = new ArrayList<>();
+        for (String nameToSplit : words) {
+            // split by uppercase letters and preserve them in the split strings
+            // also map to all lowercase
+            Set<String> split = Arrays.stream(nameToSplit.split("(?<!(^|[A-Z]))(?=[A-Z])|(?<!^)(?=[A-Z][a-z])")).map(String::toLowerCase).collect(Collectors.toSet());
```
IotaBread: This could look behind for any letter and ahead for an uppercase (i.e. `(?<=[a-zA-z])(?=[A-Z])`), instead of a lowercase behind and an uppercase ahead (`(?<!(^|[A-Z]))(?=[A-Z])`) or an uppercase and a lowercase ahead (`((?<!^)(?=[A-Z][a-z])`)

IotaBread (approved):
Looks good so far! There are a couple things in `SpellingChecker` that should be fixed though (left comments there)

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: I'll experiment, I remember having a lot of trouble picking the right regex here

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: that could definitely be changed
one thing to watch for is that I wanted to avoid splitting by apostrophe, so I'll add that to the new one

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: could be a little tricky, but I'll look into it

IotaBread: I think just passing the used word as a parameter and storing it on a different variable would be enough

buildSrc/src/main/java/quilt/internal/tasks/lint/MappingLintTask.java
eristhea: two methods are annotated with it (`getWorkerExecutor()` and the `LintAction` constructor), is it fine to just remove it?

IotaBread: I didn't see those, my bad

buildSrc/src/main/java/quilt/internal/tasks/lint/SpellingChecker.java
eristhea: yours just works :p, feeling a little silly for making something so ridiculous

IotaBread: Nah, don't worry, happens every time with regex :P

mappings/net/minecraft/block/SculkCatalystBlock.mapping
```diff
@@ -1,6 +1,6 @@
 CLASS net/minecraft/unmapped/C_qqcyyscl net/minecraft/block/SculkCatalystBlock
 	FIELD f_kqttjvnj BLOOM Lnet/minecraft/unmapped/C_xhwijdsd;
-	FIELD f_robxwiws PLUSE_TICK_DELAY I
+	FIELD f_robxwiws PLUS_TICK_DELAY I
```
IotaBread: This was probably `PULSE_TICK_DELAY`

eristhea: good catch, fixed!

PR #270, opened 2022-11-09
Do many many mappings

NoComment1105:
I should not have started on a Tuesday

PR #305, opened 2022-12-08
improve networking coverage

eristhea:
coverage!

notes:
replaces a lot of #270 due to my personal dislike of the names proposed there
moves a few things into the `net.minecraft.networking.message` package that were previously in `net.minecraft.networking` for consistency
fixes a packet being in `packet.s2c` instead of `network.packet.s2c`
refactors a few fields for consistency
refactors `getPlayer()` in `ServerCommandSource` to `getPlayerOrThrow()` since it throws an exception when a player is not found. the new `getPlayer()` returns null if not found.


OroArmor (approved):

NoComment1105 (approved):
Lgtm

PR #306, opened 2022-12-08
miscellaneous mappings, mostly dfu coverage

eristhea:
mapping dfu is so calming

notes:
fixes a bad name, one fix was marked as a schema

### DIFF
--- a/mappings/net/minecraft/item/map/MapState.mapping
+++ b/mappings/net/minecraft/item/map/MapState.mapping
@@ -3,7 +3,7 @@ CLASS net/minecraft/unmapped/C_nvpllgmg net/minecraft/item/map/MapState
 		COMMENT The banner markers to track in world.
 		COMMENT <p>
 		COMMENT Empty for the client.
-	FIELD f_fcobggon centerY I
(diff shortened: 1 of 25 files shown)


mappings/net/minecraft/item/map/MapState.mapping
```diff
@@ -3,6 +3,7 @@ CLASS net/minecraft/unmapped/C_nvpllgmg net/minecraft/item/map/MapState
 		COMMENT The banner markers to track in world.
 		COMMENT <p>
 		COMMENT Empty for the client.
+	FIELD f_fcobggon centerY I
```
violet-sh: This should be `centerZ` not `centerY`.

violet-sh (changes requested):

mappings/net/minecraft/item/map/MapState.mapping
eristhea: my mistake, sorry!

IotaBread (approved):

OroArmor (approved):

PR #307, opened 2022-12-08
Command argument mappings (and cooking categories)

Cypher121:
All the things needed for 1.19.3 QKL, mostly brigadier stuff

Not too sure about the getAsync name for messages, open to alternatives

PR #309, opened 2022-12-08
Just a basis for chat mappings

Gaming32:
This contains just a few basic chat mappings that I had time for.


eristhea:
I don't really understand why this is needed?
chat mappings are already being covered by #305 and #307

eristhea:
merged into #305, thank you!

PR #310, opened 2022-12-10
Fix #304

### LINKED ISSUE #304: Unmapped classes get mapped to hashed in the named publication
If a class has has no mappings, a hashed mapping gets marked as a named mapping in the intermediary-v2 publication.

For example, the hashed names here are part of the `named` namespace:
```
c    net/minecraft/class_459$class_460$1    net/minecraft/unmapped/C_aarntxpx$C_mxrdfjwd$C_hvkwffju
    f    Lnet/minecraft/class_459$class_460;    field_33831    f_xluscmud
```

This prevents layering mappings on top of QM. 

<details>
<summary>Relevant screenshots</summary>

QM source: 

Decompiled Minecraft source where Mojmap is layered on QM to fill in unmapped names: 

</details>

Gaming32:
The code I removed said "required by unpick". However, I diffed the `definitions.unpick` from this fork and the `definitions.unpick` from `1.19.3+build.2`, and there were no differences, so something may have changed since this code was written two months ago.

### DIFF
--- a/buildSrc/src/main/java/quilt/internal/tasks/build/AbstractTinyMergeTask.java
+++ b/buildSrc/src/main/java/quilt/internal/tasks/build/AbstractTinyMergeTask.java
@@ -71,7 +71,7 @@ public abstract class AbstractTinyMergeTask extends DefaultMappingsTask {
 
     @Internal
     protected Map<String, String> getNameAlternatives() {
-        return Collections.singletonMap("named", this.fillName); // Fill unnamed classes with hashed; Needed for remapUnpickDefinitions and possibly other things
+        return Collections.emptyMap();
     }
 
     protected MappingVisitor getFirstVisitor(MappingVisitor next) {


Gaming32:
Ok so slight issue, which isn't exactly my fault. `C_bxcgdrgm` still got mapped to hashed because it has things inside it which are named. However, it tries to access its anonymous inner class (`C_bxcgdrgm$1` on hashed), which is not remapped to hashed, and is thus not in `net.minecraft.unmapped` (in this particular case, its inner class got remapped to `com.mojang.realmsclient.gui.task.RepeatedDelayStrategy$1` because of Mojmap layering).

This is caused at its core by Tiny not supporting unmapped classes with partially mapped contents. I'd also argue that this is worth its own issue after this PR is merged, however, if the PR is merged with this outstanding issue, it would break everything.

I'm going to draft this PR for now.

eristhea:
superseded by #330

PR #312, opened 2022-12-13
remove "worldgen" from carver and feature registry names

eristhea:
So yes, this is my fault. I should have spent more time checking over the registry names.

issue:
`FEATURE` and `CARVER` registries got a `_WORLDGEN` suffix automapped, since the strings in their registry keys are `worldgen/feature` and `worldgen/carver`. I must have not been thinking when I established the mappings for those two. This PR fixes those names, so that they correspond with their `RegistryKeys` equivalents and are more concise.


OroArmor (approved):

mappings/net/minecraft/registry/Registries.mapping
```diff
@@ -23,7 +23,7 @@ CLASS net/minecraft/unmapped/C_nusqeapl net/minecraft/registry/Registries
 	FIELD f_hmxbehsm CUSTOM_STAT Lnet/minecraft/unmapped/C_tqxyjqsk;
 	FIELD f_iaquflcq MEMORY_MODULE_TYPE Lnet/minecraft/unmapped/C_zogerkic;
 	FIELD f_idnjesqk ROOT_REGISTRY_KEY Lnet/minecraft/unmapped/C_ncpywfca;
-	FIELD f_ipaoqeaf FEATURE_WORLDGEN Lnet/minecraft/unmapped/C_tqxyjqsk;
+	FIELD f_ipaoqeaf FEATURE Lnet/minecraft/unmapped/C_tqxyjqsk;
```
IotaBread, line 26: In this case it's clearer to use `WORLDGEN_FEATURE` rather than just `FEATURE` imo

IotaBread (approved):

mappings/net/minecraft/registry/Registries.mapping
eristhea: true, it's possible it could cause confusion with people thinking they have to register any sort of feature they add to the game

NoComment1105 (approved):

PR #316, opened 2022-12-17
miscellaneous new mappings

eristhea:
I was really on a roll for a while with my clever one-liners but now I'm out of ideas

notes:
has a few awkward names, I have to take another look at some of the constants


mappings/net/minecraft/block/entity/MobSpawnerBlockEntity.mapping
```diff
@@ -13,4 +13,7 @@ CLASS net/minecraft/unmapped/C_qtiirvxr net/minecraft/block/entity/MobSpawnerBlo
 		ARG 1 pos
 		ARG 2 state
 		ARG 3 blockEntity
+	METHOD m_nqofrqwu setSpawnedEntity (Lnet/minecraft/unmapped/C_ogavsvbr;Lnet/minecraft/unmapped/C_rlomrsco;)V
```
OroArmor:
```suggestion
	METHOD m_nqofrqwu setSpawnerEntity (Lnet/minecraft/unmapped/C_ogavsvbr;Lnet/minecraft/unmapped/C_rlomrsco;)V
```
maybe?

mappings/net/minecraft/client/item/TooltipContext.mapping
```diff
@@ -1,4 +1,14 @@
 CLASS net/minecraft/unmapped/C_taebrtdw net/minecraft/client/item/TooltipContext
+	FIELD f_yceowjjc NON_ADVANCED_NON_CREATIVE Lnet/minecraft/unmapped/C_taebrtdw$C_vsiqutyg;
+	FIELD f_zuchejxz ADVANCED_NON_CREATIVE Lnet/minecraft/unmapped/C_taebrtdw$C_vsiqutyg;
 	METHOD m_ecuysqaa isAdvanced ()Z
+	METHOD m_wmqdlnmf isCreative ()Z
```
OroArmor:
```suggestion
	METHOD m_wmqdlnmf isInCreativeTab ()Z
```
Not 100% sure but i think this is for when and only when the stack is in the tab, so it should be a bit clearer

OroArmor (dismissed):

mappings/net/minecraft/block/entity/MobSpawnerBlockEntity.mapping
eristhea: I considered this, but thought it might be interpreted as "set the block entity of the spawner"

OroArmor: I see. I think the fact that it takes an Entity and not a BlockEntity would be enough to know the difference

eristhea: sure, I'll go with `setSpawnerEntity` then

mappings/net/minecraft/client/item/TooltipContext.mapping
eristhea: that seems right based on a quick investigation, will commit the suggestion!

OroArmor (approved):

mappings/net/minecraft/client/item/TooltipContext.mapping
```diff
@@ -1,4 +1,14 @@
 CLASS net/minecraft/unmapped/C_taebrtdw net/minecraft/client/item/TooltipContext
```
EnnuiL: Toes stepped by #336, which yeah, I believe it has better names

mappings/net/minecraft/block/entity/MobSpawnerBlockEntity.mapping
```diff
@@ -13,4 +13,7 @@ CLASS net/minecraft/unmapped/C_qtiirvxr net/minecraft/block/entity/MobSpawnerBlo
 		ARG 1 pos
```
EnnuiL, line 13: hm, interesting point for a later PR: this class uses `MobSpawner` (the block entity's ID), but the block itself uses `Spawner` (the block/item's ID); I wonder if a convergence to `MonsterSpawner` would be better in the end;

You can ignore this message :P

mappings/net/minecraft/potion/PotionUtil.mapping
```diff
@@ -8,6 +8,10 @@ CLASS net/minecraft/unmapped/C_lyeejpyw net/minecraft/potion/PotionUtil
 		ARG 0 potion
 	METHOD m_cxismorv getColor (Ljava/util/Collection;)I
 		ARG 0 effects
+	METHOD m_dayagjta appendTooltip (Ljava/util/List;Ljava/util/List;F)V
```
EnnuiL: Toes stepped by #336, where `appendDetailsToTooltip` was used; hm, i feel like the more verbose one might be better but yeah, am not 100% sure

EnnuiL (changes requested):

mappings/net/minecraft/item/SuspiciousStewItem.mapping
```diff
@@ -2,6 +2,10 @@ CLASS net/minecraft/unmapped/C_kkuffgzc net/minecraft/item/SuspiciousStewItem
 	FIELD f_lrwjllbv EFFECT_ID_KEY Ljava/lang/String;
 	FIELD f_ssywxyld EFFECT_DURATION_KEY Ljava/lang/String;
 	FIELD f_wopcliea EFFECTS_KEY Ljava/lang/String;
+	METHOD m_cryeecjp consumeStatusEffects (Lnet/minecraft/unmapped/C_sddaxwyk;Ljava/util/function/Consumer;)V
+		ARG 0 stack
+		ARG 1 consumer
+			COMMENT handles the usage of each new status effect instance after it's created
```
EnnuiL: oh, tiny nitpick;
```suggestion
			COMMENT handles the usage of each new status effect instance after it is created
```
great usage of javadocs in order to prevent a `getStatusEffectsButPutItOnAList`! :P

EnnuiL (changes requested):
oh, one nitpick

mappings/net/minecraft/potion/PotionUtil.mapping
eristhea: just going to throw out tooltip stuff here

EnnuiL (approved):

PR #320, opened 2022-12-19
disable "inconsistent inner class entries" warning in quiltflower

eristhea:
this option to show this warning being enabled creates ~1700 lines of nonsense in the logs, it's best to just suppress it. This will make logs, especially actions logs, much more readable.

### DIFF
--- a/buildSrc/src/main/java/quilt/internal/decompile/quiltflower/QuiltflowerDecompiler.java
+++ b/buildSrc/src/main/java/quilt/internal/decompile/quiltflower/QuiltflowerDecompiler.java
@@ -6,6 +6,7 @@ import org.gradle.api.logging.LogLevel;
 import org.jetbrains.java.decompiler.main.decompiler.BaseDecompiler;
 import org.jetbrains.java.decompiler.main.extern.IBytecodeProvider;
 import org.jetbrains.java.decompiler.main.extern.IFernflowerLogger;
+import org.jetbrains.java.decompiler.main.extern.IFernflowerPreferences;
 import org.jetbrains.java.decompiler.main.extern.IResultSaver;
 import quilt.internal.decompile.AbstractDecompiler;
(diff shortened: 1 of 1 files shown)


OroArmor (approved):

buildSrc/src/main/java/quilt/internal/decompile/quiltflower/QuiltflowerDecompiler.java
```diff
@@ -37,6 +38,10 @@ public QuiltflowerDecompiler(Project project) {
     public void decompile(File file, File outputDir, Map<String, Object> options, Collection<File> libraries) {
         Path outputPath = outputDir.toPath();
 
+        // disable "inconsistent inner class" warning due to spam in the logs
+        options.put(IFernflowerPreferences.WARN_INCONSISTENT_INNER_CLASSES, "0");
+        // note: "inconsistent generic signature" warnings should also be suppressed, but that isn't currently option in quiltflower
```
IotaBread, line 43: We could suppress these by filtering them out in `LoggerImpl#writeMessage` below

IotaBread (approved):

buildSrc/src/main/java/quilt/internal/decompile/quiltflower/QuiltflowerDecompiler.java
eristhea: true, but I think it would be better to not bandaid it and add an option to quiltflower
we're using bleeding-edge quiltflower anyway!

eristhea:
update: I added a quiltflower option to disable the other warning (`inconsistent generic signature`), in https://github.com/QuiltMC/quiltflower/pull/233. We may want to wait on merging this until that's merged and we can use it.

eristhea:
given that the other exception is fixable elsewhere, I think it's fine to merge and just make a follow-up to apply the fix in either draftsman or quiltflower

PR #349, opened 2023-01-27
Update Quilt Meta CI

Akarys42:
This commit removes the HTTP request task and replaces it by the new GitHub Actions setup.

Draft until the last secret is added to the organisation.


OroArmor:
Has this been tested with another repository or manual action?

OroArmor (approved):

IotaBread (approved):

Akarys42:
> Has this been tested with another repository or manual action?

The action has been confirmed successful with various test runs done on its own repo. The only thing that I'm unsure about is whenever I'll need to publish the action or not, the docs really aren't clear about that, but even if that's the case no further action will be required in this repo.

tl;dr: even if there is breakage, it is made so the content in the various repositories don't need changing.

PR #356, opened 2023-02-07
Command plus random

OroArmor:
This will be after #355


mappings/net/minecraft/command/CommandSource.mapping
```diff
@@ -28,23 +30,31 @@ CLASS net/minecraft/unmapped/C_sasfholk net/minecraft/command/CommandSource
 		ARG 0 string
 		ARG 1 candidates
 		ARG 2 builder
+		ARG 3 isPosition
 	METHOD m_jgssjzjq suggestColumnPositions (Ljava/lang/String;Ljava/util/Collection;Lcom/mojang/brigadier/suggestion/SuggestionsBuilder;Ljava/util/function/Predicate;)Ljava/util/concurrent/CompletableFuture;
 		ARG 0 string
 		ARG 1 relativePosition
 		ARG 2 builder
+		ARG 3 isPosition
```
eristhea, line 38: this parameter doesn't really make sense to me without looking at the method's code: if I'm already passing in a parameter called `relativePosition`, why is there another called `isPosition`?

mappings/net/minecraft/command/argument/RegistryEntryOrTagArgument.mapping
```diff
@@ -0,0 +1,39 @@
+CLASS net/minecraft/unmapped/C_bwiwqfgl net/minecraft/command/argument/RegistryEntryOrTagArgument
```
eristhea, line 1: I don't like this name, but also can't think of a better way to express it
same applies to `RegistryKeyOrTagKeyResult`

eristhea (approved):
brilliant!

mappings/net/minecraft/client/gui/screen/narration/ScreenNarrator.mapping
```diff
@@ -11,7 +11,7 @@ CLASS net/minecraft/unmapped/C_yviluxux net/minecraft/client/gui/screen/narratio
 		ARG 3 partIndex
 		ARG 4 message
 	METHOD m_ugxbjsky buildNarrations (Ljava/util/function/Consumer;)V
-		ARG 1 builderConsumer
+		ARG 1 consumer
```
OroArmor:
```suggestion
		ARG 1 builderConsumer
```

mappings/net/minecraft/command/CommandSource.mapping
OroArmor: its a predicate

eristhea: ah
I guess I just assumed boolean

EnnuiL (approved):

PR #317, opened 2022-12-17
fix javadoc and field name warnings

eristhea:
redo of #302 since I messed up a git merge

here's that with a clean diff!

PR #357, opened 2023-02-08
cause build failure on bad casing

eristhea:
simple change, after #317 I feel like we should hard fail on these to ensure bad names don't get in and have to be refactored later.


EnnuiL (approved):
huhhhh those weren't errors??

OroArmor (approved):

PR #359, opened 2023-02-09
Mappings for damage type tags (& Provider) & record class

Mysterious-Dev:
All is in the title ^^.


eristhea (commented):
why `DamageTypeTagsProvider` over `DamageTypeTagProvider`?

OroArmor:
Currently tag providers dont have a consistent naming scheme. Some are `Tag` while others are `Tags`

OroArmor (approved):

eristhea (approved):
I'll approve then, but I think that should be fixed in the future

Mysterious-Dev:
I moved DamageTypeTagsProvider to `net.minecraft.data.server.tag` (instead of `net.minecraft.data.server`).

PR #366, opened 2023-02-20
import mojang's math mappings

eristhea:
math being unmapped has bothered me for too long


mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -57,10 +80,21 @@ CLASS net/minecraft/unmapped/C_zpccxxiz net/minecraft/util/math/MathHelper
...
+		ARG 1 start
+		ARG 2 end
 	METHOD m_fdvuyfhh lerpAngleDegrees (FFF)F
 		ARG 0 delta
 		ARG 1 start
 		ARG 2 end
+	METHOD m_fpdoqovb getInt (Ljava/lang/String;I)I
+		ARG 0 string
+		ARG 1 fallback
+	METHOD m_fwvsrufh rayIntersectsAABB (Lnet/minecraft/unmapped/C_vgpupfxx;Lnet/minecraft/unmapped/C_vgpupfxx;Lnet/minecraft/unmapped/C_hbcjzgoe;)Z
```
EnnuiL: because then you get bad names like AABB sneaking in ;-;

EnnuiL (changes requested):
i didn't want to simply take mojang names ;-;

mappings/net/minecraft/util/math/MathHelper.mapping
eristhea: that's the only name that bothered me out of the ones I copied

mappings/net/minecraft/util/math/MathHelper.mapping
```diff
@@ -3,18 +3,26 @@ CLASS net/minecraft/unmapped/C_zpccxxiz net/minecraft/util/math/MathHelper
 	FIELD f_aomqjoqj RADIANS_PER_DEGREE F
 	FIELD f_bciswahe DEGREES_TO_SINE_TABLE_INDEX F
 	FIELD f_buotfbwr SINE_TABLE [F
-	FIELD f_clwgxcor HALF_PI_RADIANS_SINE_TABLE_INDEX J
```
EnnuiL, line 6: oh, i'm actually curious about this field; does the new name match it's current usages?

EnnuiL (changes requested):

mappings/net/minecraft/util/math/MathHelper.mapping
eristhea: as far as I've found, its only usage is in `randomUuid`, so yes
the reason for the incorrect yarn name is because a float of the same value is used in `cos`

EnnuiL: ohh, I see; In that case? *pats mojmap*

EnnuiL (approved):
I feel like more eyes might be needed than an on-break mine who can't dig too deep :P

OroArmor (approved):

supersaiyansubtlety (approved):

PR #367, opened 2023-02-20
Pack Metadata Provider mappings

Mysterious-Dev:
All is in the title.


EnnuiL (approved):
I was worried by another PR already covering this, but it doesn't look like the case; Anyway, lgtm!

eristhea (approved):
hm, I'll have to make a pass over this later

PR #330, opened 2023-01-02
Mitigate hashed names showing up in the intermediary publication

IotaBread:
Due to a fundamental issue with Tiny V2's structure, I think this is as far as we can get on fixing this for the moment.

As opposed to the Enigma format, which stores inner classes in a tree structure, where the names of the child classes are independent on the parent ones, the Tiny format stores all classes at root-level, and their names are dependent on their parent classes. A similar issue happens with fields and methods

<details>
<summary>An example</summary>

- We have class `a`, with an inner class `b`, which has inner class `c`
- We also have class `d`, with inner classes `e` and `f`
- `a`, `b`, `d` and `e` don't have names
- `c` is named `C_Class`, `f` is `F_Class`

In the Enigma format:
- Class declarations follow the format `CLASS <obf-name> [<deobf-name>]`
- Inner class declarations are "inside" their parent
- `a` has declaration `CLASS a`
  - `b` has declaration `CLASS b`
    - `c` has declaration `CLASS c C_Class`
- `d` has declaration `CLASS d`
  - `e` has declaration `CLASS e`
  - `f` has declaration `CLASS f F_Class`

On the other hand, in the Tiny V2 format:
- Class declarations follow the format `c <full-obf-name> [<full-deobf-name>]`
- All class declarations are at root level
- `a` has declaration `c a`
  - `b` has declaration `c a$b a$b`
    - `c` has declaration `c a$b$c a$b$C_Class`
- `d` has declaration `c d d`
  - `e` has declaration `c d$e`
  - `f` has declaration `c d$f d$F_Class`

The main difference is that Tiny uses full binary (or qualified?) names for it's classes, while Enigma uses names per entry. Notice how in Tiny, `a`, `b` and `d` have "deobf" names, while in Enigma they don't

</details>

Because of this problem, you may find a class named `C_abcdefgh` instead of `class_1234`, because it has an inner class named `FooBar`. You may also find `m_abcdefgh` instead of `method_12345` because it has a parameter named `world`

PR #371, opened 2023-02-23
Remove as much hashed names as possible from intermediary publication 

### LINKED ISSUE #334: `C_cifoaixd$C_lcybgaxw$C_amqftyfx$C_iinezlaz` is broken
Its in ChunkBuilder.BuiltChunk.RebuildTask. I found it by that class' render method, its return value is that class `C_cifoaixd$C_lcybgaxw$C_amqftyfx$C_iinezlaz` and it is very broken. It wont let me import it. `C_cifoaixd$C_lcybgaxw$C_amqftyfx` (the first part of it) just maps to `ChunkBuilder.BuiltChunk.RebuildTask`. I dont know what else to put here.

IotaBread:
Depends on #370, replaces #330. Better reviewed by filtering out commits from #370. Doesn't fix #334 yet

- Removes `buildTinyWithEnum` task, as it was broken and used just by `remapUnpickDefinitions`
  - Replaced by `mergeTinyV2`, which includes the changes from `insertAutoGeneratedMappings`
- Tiny merge tasks no longer fill namespaces
- Unobfuscated methods now keep their name, even when `intermediary` would override them
- `remapUnpickDefinitions` fills the namespace it needs during runtime, and doesn't apply those changes to a file, which would add unnecessary mappings
- Added a bunch of buildscript tests for these changes

Can't be merged until #334 is fixed


IotaBread:
Alright, the pr should be good to review. Here's a quick summary of the changes this introduces:
- Unmapped methods and fields should no longer have hashed names within the intermediary publication.
- Completely unmapped classes shouldn't contain any hashed names either.
- Due to the structure of the Tiny V2 mappings format, unmapped classes with at least one entry or comment in them will use their hashed name
  - This is required to fix remapping issues with inner classes of unmapped ones.
  - This isn't needed for unmapped classes without any named inner class, but I haven't found a good way to solve this, so it's going to have to be like this for now
  - The Enigma Tiny V2 writer also saves unmapped classes as named with their hashed names, so some extra work is needed there to fix this
    - > Most of the time this isn't an issue since most classes in the tree have a mapping, but it does become an issue when the class, without a mapping, has a non-class child in the tree

OroArmor (approved):

buildSrc/src/main/java/quilt/internal/mappingio/UnmappedNameRemoverVisitor.java
```diff
@@ -12,7 +12,7 @@
 
 /**
  * Removes unmapped names from the {@link #targetNamespace target namespace}. Whether a name is unmapped is determined
- * by {@link #unmappedPredicate}
+ * either if it's name on the target namespace is the same as on the {@link #unmappedNamespace unmapped one}, or by {@link #unmappedPredicate}
```
eristhea:
```suggestion
 * either if its name on the target namespace is the same as on the {@link #unmappedNamespace unmapped one}, or by {@link #unmappedPredicate}
```

eristhea (commented):
absolutely love the tests :)

eristhea (approved):

PR #376, opened 2023-02-26
even more entity things

eristhea:
when combined with #361, this pr gives the entire `entity` package decent coverage. not perfect, but decent.
I aimed to map the whole package but decided that I didn't want an unreviewable monster

- maps the new `DisplayEntity` and `InteractionEntity`
- misc


mappings/net/minecraft/entity/damage/DamageTracker.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/unmapped/C_xkhemiqp net/minecraft/entity/damage/DamageTracker
+	FIELD f_bbxvtwjo INTENTIONAL_GAME_DESIGN_STYLE Lnet/minecraft/unmapped/C_cpwnhism;
```
EnnuiL, line 2: oh my god i love this so much

mappings/net/minecraft/entity/Targeter.mapping
```diff
@@ -0,0 +1,2 @@
+CLASS net/minecraft/unmapped/C_vgutocgh net/minecraft/entity/Targeter
```
EnnuiL: hm, i genuinely wonder if it's "targetter" or "targeter"

EnnuiL (changes requested):
Ah, just one thing!

mappings/net/minecraft/entity/Targeter.mapping
eristhea: getter is the more British way I'd think

mappings/net/minecraft/entity/damage/DamageTracker.mapping
eristhea: thank mojang!

mappings/net/minecraft/entity/Targeter.mapping
EnnuiL: heck, i'm a bit british then

EnnuiL (approved):

OroArmor (approved):

mappings/net/minecraft/item/ModelTransformationMode.mapping
```diff
@@ -0,0 +1,10 @@
+CLASS net/minecraft/unmapped/C_byhguelf net/minecraft/item/ModelTransformationMode
```
OroArmor: Can we move this class to `net/minecraft/client/render/model/json`?

OroArmor (changes requested):

mappings/net/minecraft/item/ModelTransformationMode.mapping
eristhea: one concern is that I used the package to signify that it's unique to items, but I like that refactor

OroArmor (approved):

PR #373, opened 2023-02-24
1.19.4 random things

eristhea:
me when I test enigma
just trying to improve coverage before 1.19.4

- main goal of this pr is to take messy classes and complete them
- completes mappings in the `text`, `registry`, and `scoreboard` packages
- focuses a lot on the `util` package
- fixes a couple inconsistent names

PR #385, opened 2023-03-12
net.minecraft.item coverage

moehreag:
Completes coverage in the net.minecraft.item package. This is rather small since most classes were already completely mapped.

*I hope there aren't too many things that need to be changed..*

*Of course rai was faster already with a few of these changes in #373  :p*

### DIFF
--- a/mappings/net/minecraft/item/VanishableArmorItem.mapping
+++ b/mappings/net/minecraft/item/VanishableArmorItem.mapping
@@ -1,0 +1,8 @@
+CLASS net/minecraft/unmapped/C_kelbxmwg net/minecraft/item/VanishableArmorItem
+	METHOD m_bgvbfdaw get (Lnet/minecraft/unmapped/C_sddaxwyk;)Lnet/minecraft/unmapped/C_kelbxmwg;
+	METHOD m_gvcvcgqk use (Lnet/minecraft/unmapped/C_vorddnax;Lnet/minecraft/unmapped/C_cdctfzbn;Lnet/minecraft/unmapped/C_jzrpycqo;Lnet/minecraft/unmapped/C_laxmzoqs;)Lnet/minecraft/unmapped/C_upvihyur;
+		ARG 1 item
+		ARG 3 entity
+		ARG 4 hand
+	METHOD m_ndbemose getEquipSound ()Lnet/minecraft/unmapped/C_avavozay;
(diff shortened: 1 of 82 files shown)


eristhea:
> Of course rai was faster already with a few of these changes in https://github.com/QuiltMC/quilt-mappings/pull/373 :p

<3

eristhea (approved):
brilliant!

mappings/net/minecraft/item/VanishableArmorItem.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/unmapped/C_kelbxmwg net/minecraft/item/VanishableArmorItem
```
eristhea: actually one thing.
this seems like *way* too specific of a name, given that all equipment that can be worn can have vanishing curse
could we not just name this something like `Equippable` or `Equipment`?
probably the longest name I'd be ok with is `WearableEquipment`

eristhea (changes requested):

mappings/net/minecraft/item/VanishableArmorItem.mapping
OroArmor: I think this used to be `Vanishable`

OroArmor (approved):

mappings/net/minecraft/item/VanishableArmorItem.mapping
moehreag: No, but this *extends* `Vanishable` and is only used by ArmorItems, Skulls, Carved Pumpkins, Elytras and Shields (which are basically all items that can be worn).

OroArmor: Ah ok. Then i would say `WearableVanishable` or the other way around

OroArmor: I hate that so much though

eristhea: yeah no that's a bad name
I checked mojang and they simply call it `Equippable` so I think it's best we match that

moehreag: sure!

eristhea (approved):

EnnuiL (approved):

PR #392, opened 2023-03-17
mappings & refactor for tags providers

Mysterious-Dev:
All is in the title ^^.

I update the title for block tags to standardize with others.


NoComment1105 (approved):

IotaBread (approved):

EnnuiL (approved):
Hm, now I wonder if "OneTwentyUpdate" would be better than "OneTwenty" or the pretty moj-y "UpdateOneTwenty", although hm, "1.20 block/biome tags" makes more sense

PR #395, opened 2023-03-19
Use draw and not computeVertices for methods that draw in TextRenderer.

LambdAurora:
`draw` is much more intuitive than `compute` in this context.


EnnuiL (approved):

IotaBread (approved):

eristhea (approved):
I like how nearly all of your prs fix an incorrectly formatted \<p>

moehreag:
I approve. I might have been tired.

PR #397, opened 2023-03-22
a few screen handler mappings

eristhea:
@OroArmor specifically requested I map `C_vkmtnvmw`

PR #399, opened 2023-03-23
screen handler mappings, 1.19 edition

eristhea:
oro was not satisfied with #397, so I had to do this
this is a copy of #397 to be merged alongside it, please review there and I'll keep it synced!


EnnuiL:
i'm confused, it looks like the same picture as #397; maybe the target was wrong?

eristhea:
oh, my mistake
I'll fix the target

EnnuiL (approved):

Mysterious-Dev:
That's weird, why the target branch is for 1.20 version ?

NoComment1105:
Maybe i shouldn't have update based, two seconds please

NoComment1105:
Sorted, thankfully there were no conflicts

PR #406, opened 2023-04-12
Bunch of stuff: signs, items, loot tables, and more.

LambdAurora:


EnnuiL (approved):

mappings/net/minecraft/util/CommonColors.mapping
```diff
@@ -0,0 +1,8 @@
+CLASS net/minecraft/unmapped/C_oauvuath net/minecraft/util/CommonColors
```
eristhea, line 1: `Common` is a bit of a mojmap thing, I think we could get away with just `Colors` here

eristhea (approved):
lovely!

mappings/net/minecraft/util/CommonColors.mapping
EnnuiL: i actually was thinking of like, considering adopting that name; screentext after all isn't really a better name than commontext (commoncomponents)

PR #409, opened 2023-04-22
map GuiGraphics and friends (formerly DrawableHelper)

eristhea:
aims to restore gui rendering back to its 23w14a days

the change to `GuiGraphics` is motivated by the fact that it now no longer is comprised of static methods, but is now an argument passed around to various GUI rendering methods.

notes:
- fixes some args that were somehow scrambled in `TooltipPositioner`


eristhea:
marked as draft until I map all the methods that now take it as a parameter

EnnuiL (approved):

OroArmor (approved):

PR #414, opened 2023-04-29
spellcheck improvements

eristhea:
- update dictionary revision (remove "behaviour" and "bonemeal")
- update minecraft_specific_words.txt (remove "bonemeal")
- fix checking all-uppercase words
- fix new issues introduced by those changes
- store dictionary file based on revision (allows having different revisions on different git branches)
   - previously, if you swapped the dictionary revision it wouldn't update, since the file already existed. this also meant that if you swapped the revision and deleted the file, forcing an update, then switched branches to one with a different revision, you would have the wrong revision, leading to build failures.

notes:
should we handle deleting old dictionaries? each one is <5 mb, and I expect updates to be less than monthly. I don't think this will cause too many disk space concerns, especially since gradle caches need to get cleared a lot anyway.


EnnuiL (approved):

mappings/net/minecraft/item/SmithingTemplateItem.mapping
```diff
@@ -13,7 +13,7 @@ CLASS net/minecraft/unmapped/C_hkwshalt net/minecraft/item/SmithingTemplateItem
 	FIELD f_iuzwfkby NETHERITE_UPGRADE_TEXT Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_kanabeos baseIcons Ljava/util/List;
 	FIELD f_kflahqzy BLUE Lnet/minecraft/unmapped/C_tnezalvh;
-	FIELD f_lwrbmpyz HELMENT_ICON Lnet/minecraft/unmapped/C_ncpywfca;
```
IotaBread, line 16: hemlent

IotaBread (approved):

mappings/net/minecraft/item/SmithingTemplateItem.mapping
IotaBread: lmao I mispelled the mispelling

eristhea: HELP

PR #416, opened 2023-05-08
rename intprovider and floatprovider

### LINKED ISSUE #410: refactor all packages to use underscores instead of two consecutive words
things like `net/minecraft/util/intprovider` are bad.

todo:
- [ ] add package names to spell check
- [ ] fix all issues

(waiting on #409 before I work on this)

eristhea:
fixes #410

### DIFF
--- a/buildSrc/src/main/resources/minecraft_specific_words.txt
+++ b/buildSrc/src/main/resources/minecraft_specific_words.txt
@@ -389,8 +389,6 @@ nano
 
 // functions
 alloc
-floatprovider
-intprovider
 println
 raycast
 uniquifier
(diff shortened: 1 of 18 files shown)


OroArmor (approved):

IotaBread (approved):

EnnuiL (changes requested):
While I like those new names better, I do wonder if something like `providers/<type>` would be better

LambdAurora:
Not allowed due to them being primitives.

EnnuiL:
> Not allowed due to them being primitives.

java what in the world

EnnuiL (approved):
I accept defeat by Java then; This name is much much better than gluecase anyway

LambdAurora (commented):
Definitely feeling a bit weird about using underscores.
But our conventions state that we should follow a specific structure and respect [Java conventions](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html) which doesn't seem to disallow underscores.

But it does indicate that if you want to use `int` as a single component it should be written as `int_` which is quite worse.

eristhea:
the arguments kinda boil down to two main points from what I've seen
- for: the separation of the words makes things more clear
- against: the underscores feel strange to people who use java a lot

I think that since our goal is readability and not familiarity, underscores are what we should go for

PR #424, opened 2023-05-15
Map tag stuff and some registry stuff.

LambdAurora:
Would help porting QSL Tags API.


EnnuiL (approved):
`CREATIVE_MODE_TAB` looks a bit weird considering we avoided that name in particular in favor of item groups, but oh well, I'll let it go

LambdAurora:
The name was the same in RegistryKeys, though that one was auto-mapped, always can change that

OroArmor (approved):

eristhea (approved):

EnnuiL:
> The name was the same in RegistryKeys, though that one was auto-mapped, always can change that

oh, fair enough

IotaBread (approved):

PR #425, opened 2023-05-18
loot tables and entities

eristhea:
just general poking around and mapping

### DIFF
--- a/mappings/net/minecraft/entity/Tameable.mapping
+++ b/mappings/net/minecraft/entity/Tameable.mapping
@@ -1,3 +1,4 @@
 CLASS net/minecraft/unmapped/C_lkirduch net/minecraft/entity/Tameable
+	METHOD m_bllryzha getEntityView ()Lnet/minecraft/unmapped/C_nyckmiua;
 	METHOD m_jaxlomko getOwner ()Lnet/minecraft/unmapped/C_usxaxydn;
 	METHOD m_syyylunm getOwnerUuid ()Ljava/util/UUID;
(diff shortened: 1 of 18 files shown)


EnnuiL:
hm, it appears that you forgot to undraft this PR

mappings/net/minecraft/entity/Tameable.mapping
```diff
@@ -1,3 +1,4 @@
 CLASS net/minecraft/unmapped/C_lkirduch net/minecraft/entity/Tameable
+	METHOD m_bllryzha getEntityView ()Lnet/minecraft/unmapped/C_nyckmiua;
```
EnnuiL, line 2: get entity view from who? i unfortunately couldn't confirm it myself when i strumbled on it, but i suspect it is the owner's entity view

EnnuiL (changes requested):
other than the 1.20-pre5 merger fail and the following poke, this pr looks good!

mappings/net/minecraft/entity/Tameable.mapping
eristhea: good catch, I'll double check and specify

eristhea: purely based on the implementation of `getOwner`, I think the entity view belongs to the original entity
despite not being able to locate implementations

EnnuiL: ah i see

EnnuiL (approved):

EnnuiL:
aw, it'll miss 1.20-pre5 merge

supersaiyansubtlety (approved):

EnnuiL:
oh nvm

PR #437, opened 2023-05-29
Adjust the FCP durations for QM's needs

EnnuiL:
While this also makes the FCPs match the reality, I also propose the following:
 - FCPs for `t: new` and `t: documentation` can be fast-tracked to 1 day if there are 3 approvals *and* it targets a snapshot
 - FCPs for `t: refactor` has been shortened to 2 days due to how FCPs longer than 3 days have been hurting us, and plus the requirement for 2 "core" approvals *and* 3 approvals in general have already inflated the time it takes for such a PR to be approved anyway, with the 2 "core" approvals already making them go through higher scrutiny than usual
   - Update: `t: medium` and `t: large` refactors, however, will respectively have 3 and 4 days of FCP, with the expectation being that the PR indeed is a massive refactoring PR (this ideally means that a one-line refactor in the middle of a `t: new` and `t: medium` PR does not count for bumping the days)
 - FCPs for `t: toolchain` now require 2 approvals, although now that the mappings team is big enough for that to be constantly met, I believe that is the standard nowadays (and plus, toolchain PRs tend to be of the kind that doesn't benefit much from a fast iteration mentality)
 - The Ennui Manuever™ has been ported to Quilt Mappings, allowing the Mappings team to alter individual FCPs if there is a consensus
 
A FCP policy for Quilt Mappings must emphasize the need of fast iteration, especially for a project that needs it, and we can't afford FCPs longer than 2 days at this moment


IotaBread (approved):

CONTRIBUTING.md
```diff
@@ -72,58 +72,52 @@ with [RFC 39](https://github.com/QuiltMC/rfcs/blob/master/structure/0039-pr-poli
...
 
 **Required Approvals**: 3
-- At least 2 approvals must come directly from a Mappings team member, and the other one can come from a Mappings
-Triage member or another Mappings Team member.
+- At least 2 approvals must come directly from a Mappings team member.
 
-**Final Comment Period**:
-- Snapshots: 3 days
-- Releases: 6 days
+**Final Comment Period**: 2 days
```
LambdAurora: I'd add something saying that if the PR is really big the FCP period can be longer as seen fit.

EnnuiL: Ah sure, I can introduce a mechanism for that; While it's a bit tempting to rely on the PR being more than `t: medium`, yeah, I might make this like, something to be decided manually

eristhea: I would personally bump it up one day for every stage above small: 3 for medium, 4 for large

CONTRIBUTING.md
```diff
@@ -72,58 +72,52 @@ with [RFC 39](https://github.com/QuiltMC/rfcs/blob/master/structure/0039-pr-poli
...
 
 ### `T: documentation`
 
 **Description**: Used for pull requests that focus mainly on adding or changing documentation.
 
 **Required Approvals**: 2
-- At least 1 approval must come directly from a Mappings team member, and the other one can come from a Mappings
-Triage member or another Mappings Team member.
+- At least 1 approval must come directly from a Mappings team member.
```
eristhea, line 104: I don't like having a single point here, I think we have a big enough team that we can still say 2 reviews are necessary

CONTRIBUTING.md
```diff
@@ -72,58 +72,52 @@ with [RFC 39](https://github.com/QuiltMC/rfcs/blob/master/structure/0039-pr-poli
...
 
 **Final Comment Period**:
-- Snapshots: 2 days
-- Releases: 4 days
+- Snapshots: 2 days (1 day with 3+ approvals)
+- Releases: 2 days
```
eristhea: just a note, these can merged since they're identical

CONTRIBUTING.md
EnnuiL: I'm confused here; the intention here is to make it clear that only 1 review from the Mappings team is required while the rest can come from the Mappings Triage team, I didn't change the number of approvals here

CONTRIBUTING.md
EnnuiL: wdym?

CONTRIBUTING.md
EnnuiL: ah, I agree with that

CONTRIBUTING.md
eristhea: ah I misinpreted that as eliminating the requirement for a triage review

CONTRIBUTING.md
eristhea: `2 days (1 day with 3+ approvals for snapshots)`
current format is fine though

eristhea (approved):

CONTRIBUTING.md
EnnuiL: ah, I see

CONTRIBUTING.md
```diff
@@ -72,64 +72,55 @@ with [RFC 39](https://github.com/QuiltMC/rfcs/blob/master/structure/0039-pr-poli
...
-- At least 2 approvals must come directly from a Mappings team member, and the other one can come from a Mappings
-Triage member or another Mappings Team member.
+- At least 2 approvals must come directly from a Mappings team member.
 
 **Final Comment Period**:
-- Snapshots: 3 days
-- Releases: 6 days
+- `T: small` and below: 2 days
+- `T: medium`: 3 days
+- `T: large` and above: 2 days
```
supersaiyansubtlety:
```suggestion
- `T: large` and above: 4 days
```
I assume this is a typo based on [this](https://github.com/QuiltMC/quilt-mappings/pull/437#discussion_r1210912642) conversation?

supersaiyansubtlety (changes requested):

CONTRIBUTING.md
EnnuiL: oh whoops

supersaiyansubtlety (approved):

CONTRIBUTING.md
```diff
@@ -72,64 +72,55 @@ with [RFC 39](https://github.com/QuiltMC/rfcs/blob/master/structure/0039-pr-poli
...
-- At least 2 approvals must come directly from a Mappings team member, and the other one can come from a Mappings
-Triage member or another Mappings Team member.
+- At least 2 approvals must come directly from a Mappings team member.
 
 **Final Comment Period**:
-- Snapshots: 3 days
-- Releases: 6 days
+- `T: small` and below: 2 days
+- `T: medium`: 3 days
+- `T: large` and above: 4 days
```
eristhea: these should be `s:` not `t:`

eristhea (changes requested):

eristhea (approved):

supersaiyansubtlety (approved):

OroArmor (approved):

LambdAurora (approved):

PR #449, opened 2023-06-14
rename placeLogsAndLeaves to updateLeavesDistances

### LINKED ISSUE #448: `TreeFeature#placeLogsAndLeaves` is a misnomer
`net.minecraft.world.gen.feature.TreeFeature#placeLogsAndLeaves` places no logs, and only replaces existing leaves states with updated distance properties.

Something like `setLeavesDistances` would be an accurate description.

supersaiyansubtlety:
Fixes #448


LambdAurora:
Shouldn't this be update instead of set? It doesn't set a given value, it updates the values based on surroundings if I understood right

supersaiyansubtlety:
> Shouldn't this be update instead of set? It doesn't set a given value, it updates the values based on surroundings if I understood right

supersaiyansubtlety:
Missclicks... sorry

supersaiyansubtlety:
> Shouldn't this be update instead of set? It doesn't set a given value, it updates the values based on surroundings if I understood right

You're right, update is better

eristhea (approved):

EnnuiL (approved):

LambdAurora (approved):

PR #454, opened 2023-06-26
*nearly* polish off `net.minecraft.server`

eristhea:
mostly completes minecraft's `server` package, and maps a bunch of stuff used by it. all deletions are fixing incorrect names.


EnnuiL (approved):

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -176,7 +176,7 @@ CLASS net/minecraft/unmapped/C_astfners net/minecraft/entity/Entity
 	METHOD m_bzrtaquy getVehicle ()Lnet/minecraft/unmapped/C_astfners;
 	METHOD m_bzveerey collides ()Z
 	METHOD m_cckfyxyf fall (DZLnet/minecraft/unmapped/C_txtbiemp;Lnet/minecraft/unmapped/C_hynzadkk;)V
-		ARG 1 heightDifference
+		ARG 1 yDifference
```
LambdAurora:
```suggestion
		ARG 1 fallDistance
```

Perhaps that'd be more accurate?

mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
```diff
@@ -110,7 +112,9 @@ CLASS net/minecraft/unmapped/C_mxrobsgg net/minecraft/server/network/ServerPlaye
 	METHOD m_gtxxzowf (Lnet/minecraft/unmapped/C_rdaqiwdt;Lnet/minecraft/unmapped/C_cpwnhism;)Lnet/minecraft/unmapped/C_cpwnhism;
 		ARG 1 style
 	METHOD m_gynondrt handleFall (DDDZ)V
-		ARG 1 heightDifference
+		ARG 1 xDifference
+		ARG 3 yDifference
+		ARG 5 zDifference
```
LambdAurora:
```suggestion
		ARG 1 xOffset
		ARG 3 fallDistance
		ARG 5 zOffset
```

Though, those names are less harmonic compared to the PRed names.

mappings/net/minecraft/server/world/ChunkLevels.mapping
```diff
@@ -0,0 +1,19 @@
+CLASS net/minecraft/unmapped/C_xuardpja net/minecraft/server/world/ChunkLevels
```
LambdAurora, line 1: I'd love some quick javadoc on the fields (with a little `{@value}` in them).

mappings/net/minecraft/server/network/ServerPlayerEntity.mapping
eristhea: I think the added clarity of `fallDistance` is worth it

mappings/net/minecraft/server/world/ChunkLevels.mapping
eristhea: will do that tonight! they're fairly self-explanatory, so I'll keep it short

mappings/net/minecraft/server/network/EntityTrackerEntry.mapping
```diff
@@ -21,14 +24,22 @@ CLASS net/minecraft/unmapped/C_yagtbmmk net/minecraft/server/network/EntityTrack
...
 		ARG 5 receiver
 	METHOD m_acqaqxfb sendPackets (Lnet/minecraft/unmapped/C_mxrobsgg;Ljava/util/function/Consumer;)V
+		ARG 1 player
 		ARG 2 sender
 	METHOD m_acwzlrjk sendSyncPacket (Lnet/minecraft/unmapped/C_zyjtrjrl;)V
 		COMMENT Sends a packet for synchronization with watcher and tracked player (if applicable)
 		ARG 1 packet
 	METHOD m_dlxmtkps syncEntityData ()V
 		COMMENT Synchronizes tracked data and attributes
+	METHOD m_eszryakz getNonUniqueEntries (Ljava/util/List;Ljava/util/List;)Ljava/util/stream/Stream;
```
supersaiyansubtlety:
```suggestion
	METHOD m_eszryakz getDifference(Ljava/util/List;Ljava/util/List;)Ljava/util/stream/Stream;
```
The stream returned contains all elements in `list2` *not* contained in `list1`, so the original name is backwards. 

Additionally, I think "unique" strongly implies that the method filters out duplicate entries, which it doesn't. 

I have two minor reservations about `getDifference` though:
1. the paramaters are `Lists`, not java `Set`s [but I think it's pretty clear how 'difference' would be applied to lists]
2. the returned stream is `list2` minus `list1`, not the more intuitive `list1` minus `list2` [but I doubt there's a concise-enough way to phrase that to put it in the name, maybe something for javadoc]

supersaiyansubtlety (changes requested):

mappings/net/minecraft/server/network/EntityTrackerEntry.mapping
eristhea: I was treating this as strictly a utility method, but I properly checked the moj and its usages and it's only used to find removed passengers of an entity. I've named it `streamRemovedPassengers` to reflect this!

supersaiyansubtlety (approved):

PR #456, opened 2023-07-01
yet more world stuff

eristhea:
polishing off more things in `net.minecraft.world` before finally going after world generation. does a minor refactor in the form of moving some util classes to a `path` package, to make things a little more organised.

### DIFF
--- a/mappings/net/minecraft/world/storage/StorageIoWorker.mapping
+++ b/mappings/net/minecraft/world/storage/StorageIoWorker.mapping
@@ -1,5 +1,7 @@
 CLASS net/minecraft/unmapped/C_ffgsaknh net/minecraft/world/storage/StorageIoWorker
+	FIELD f_cmzfptka MAX_REGION_CACHE_SIZE I
 	FIELD f_ebctbref storage Lnet/minecraft/unmapped/C_ivfseflh;
+	FIELD f_eqfuhsox blendingStatusCaches Lit/unimi/dsi/fastutil/longs/Long2ObjectLinkedOpenHashMap;
 	FIELD f_prbotxpi closed Ljava/util/concurrent/atomic/AtomicBoolean;
 	FIELD f_pyjvjyaq results Ljava/util/Map;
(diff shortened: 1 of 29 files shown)


EnnuiL (approved):

mappings/net/minecraft/world/storage/StorageIoWorker.mapping
```diff
@@ -9,29 +11,45 @@ CLASS net/minecraft/unmapped/C_ffgsaknh net/minecraft/world/storage/StorageIoWor
...
 	METHOD m_jjukvdrp completeAll (Z)Ljava/util/concurrent/CompletableFuture;
 		ARG 1 sync
+	METHOD m_jsyufhll (I)[Ljava/util/concurrent/CompletableFuture;
+		ARG 0 i
 	METHOD m_msmawqmw readChunkData (Lnet/minecraft/unmapped/C_ynrszrtu;)Ljava/util/concurrent/CompletableFuture;
 		ARG 1 pos
 	METHOD m_nghssqns run (Ljava/util/function/Supplier;)Ljava/util/concurrent/CompletableFuture;
 		ARG 1 task
-	METHOD m_nwcchuzp (II)Ljava/util/concurrent/CompletableFuture;
+	METHOD m_nwcchuzp getOrComputeBlendingStatus (II)Ljava/util/concurrent/CompletableFuture;
```
supersaiyansubtlety:
```suggestion
	METHOD m_nwcchuzp getBlendingStatus (II)Ljava/util/concurrent/CompletableFuture;
```
Seems like this falls under this [convention](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#getters-setters-withers-and-creators):
> Use "get" for non-boolean getters and other methods that calculate some property with no side effects other than caching a value in a private field.

Also `getOrComputeBlendingStatus` made me think it would act more like Java Map's `computeIfAbsent`, where a method to compute the value needs to be passed.

supersaiyansubtlety (changes requested):

supersaiyansubtlety (approved):

PR #457, opened 2023-07-01
specify that some `Text.translatable` methods take a fallback

eristhea:
I'm personally neutral on this change, but it was suggested by @patbox. discuss!


EnnuiL (approved):
This looks fine, but I do wonder how we could simplify that name; it feels a bit too verbose for the simple pattern it is going for

eristhea:
I spent a couple minutes thinking, but I couldn't find a simpler way to convey "make a translatable text with a fallback"

Patbox:
For record, both yarn and mojmap use the same name

supersaiyansubtlety (approved):

EnnuiL:
> For record, both yarn and mojmap use the same name

huh, that's nice to know

moehreag (approved):

IotaBread (approved):
Another alternative could be `translatableOr`, though I don't think it really fits

EnnuiL:
> Another alternative could be `translatableOr`, though I don't think it really fits

yeah, it looks too ambiguous

PR #458, opened 2023-07-02
Midnight happenings

EnnuiL:
midnight happenings

### DIFF
--- a/mappings/net/minecraft/client/particle/CherryLeafParticle.mapping
+++ b/mappings/net/minecraft/client/particle/CherryLeafParticle.mapping
@@ -5,4 +5,4 @@ CLASS net/minecraft/unmapped/C_ewetkmcp net/minecraft/client/particle/CherryLeaf
 		ARG 2 x
 		ARG 4 y
 		ARG 6 z
-		ARG 8 SpriteProvider
+		ARG 8 spriteProvider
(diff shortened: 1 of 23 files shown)


mappings/net/minecraft/client/particle/CherryLeafParticle.mapping
```diff
@@ -1 +1,8 @@
 CLASS net/minecraft/unmapped/C_ewetkmcp net/minecraft/client/particle/CherryLeafParticle
+	FIELD f_tmcykyqu rotationSpeed F
+	METHOD <init> (Lnet/minecraft/unmapped/C_ghdnlrrw;DDDLnet/minecraft/unmapped/C_audtngae;)V
+		ARG 1 world
+		ARG 2 x
+		ARG 4 y
+		ARG 6 z
+		ARG 8 SpriteProvider
```
IotaBread:
```suggestion
		ARG 8 spriteProvider
```

IotaBread (approved):

mappings/net/minecraft/client/particle/CherryLeafParticle.mapping
EnnuiL: yes am already on the way

supersaiyansubtlety (approved):

eristhea (changes requested):
I'm not a big fan of the word order on `velocityXYZ` parameters, I'd prefer we went with the natural `xyzVelocity`

EnnuiL:
> I'm not a big fan of the word order on `velocityXYZ` parameters, I'd prefer we went with the natural `xyzVelocity`

The problem is that it goes against [the convention](https://github.com/QuiltMC/quilt-mappings/blob/1.20.1/CONVENTIONS.md#coordinates) about that

eristhea:
I'd argue that `xyzVelocity` fits the convention better -- you wouldn't say "this updates the velocity X of the entity to 5 blocks per second", the more natural sentence is "updates the x velocity"

EnnuiL:
> I'd argue that `xyzVelocity` fits the convention better -- you wouldn't say "this updates the velocity X of the entity to 5 blocks per second", the more natural sentence is "updates the x velocity"

I'd argue that if you think of it as "velocity's X", things end making more sense; But oh well, I feel like a convention update/mass refactor would be a separate PR

eristhea:
ok, I'll put it on my todo list :p

eristhea (approved):

PR #464, opened 2023-07-15
improve consistency of Text names

moehreag:
- some fields aren't possible to be renamed due to naming conflicts (and I didn't want to change the other fields due to the created errors which would be created for mods using these fields).
- related discussion: https://discord.com/channels/817576132726620200/1104011466660659273/1125668806895882330

I hope I spotted most occurrences of this.


mappings/net/minecraft/village/raid/Raid.mapping
```diff
@@ -13,11 +13,11 @@ CLASS net/minecraft/unmapped/C_szefbyex net/minecraft/village/raid/Raid
 	FIELD f_fnxkdmas DEFEAT_TITLE Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_gknmsuhn SPAWN_ATTEMPTS I
 	FIELD f_jakhyora MAX_ACTIVE_TICKS I
-	FIELD f_jbpzyzdv VICTORY_SUFFIX_TEXT Lnet/minecraft/unmapped/C_rdaqiwdt;
+	FIELD f_jbpzyzdv VICTORY Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_kdhitzkh RAID_REMOVE_THRESHOLD_SQUARED I
 	FIELD f_kujahjos started Z
 	FIELD f_kyzgcvgr random Lnet/minecraft/unmapped/C_rlomrsco;
-	FIELD f_lqyogwfr DEFEAT_SUFFIX_TEXT Lnet/minecraft/unmapped/C_rdaqiwdt;
+	FIELD f_lqyogwfr DEFEAT_SUFFIX Lnet/minecraft/unmapped/C_rdaqiwdt;
```
supersaiyansubtlety:
```suggestion
	FIELD f_lqyogwfr DEFEAT Lnet/minecraft/unmapped/C_rdaqiwdt;
```
The format of this should definitely match `VICTORY` (whether they both have `_SUFFIX` or not) because their translation keys follow the same pattern, and I'm in favor of dropping the `_SUFFIX` suffix to follow the convention of matching the translation key.

supersaiyansubtlety (changes requested):
> some fields aren't possible to be renamed due to naming conflicts (and I didn't want to change the other fields due to the created errors which would be created for mods using these fields).

I don't think it's good idea to shy away from refactors that will require mod devs to refactor to as well.

eristhea (approved):
love that this sneaks in some new mappings as well!

eristhea:
also -- yes, we're never going to skip out on a good refactor just because it means mod devs will probably have to change things

Mysterious-Dev (approved):

supersaiyansubtlety (approved):

PR #479, opened 2023-08-12
A small amount of 1.20.2 mappings

EnnuiL:


mappings/net/minecraft/entity/attribute/EntityAttributes.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/unmapped/C_jtywurvj net/minecraft/entity/attribute/EntityAttributes
+	FIELD f_irrggfyr GENERIC_MAX_ABSORPTION Lnet/minecraft/unmapped/C_ppzfbbsy;
```
EnnuiL, line 2: Weirdly enough, the autofiller has failed here

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -241,6 +241,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 		FIELD f_jnxtuzdb hardness F
 		FIELD f_jvcpotqu isAir Z
 		FIELD f_lpessbtt pistonBehavior Lnet/minecraft/unmapped/C_mojvzhua;
+		FIELD f_lqzypovm spawnBlockDustParticles Z
```
supersaiyansubtlety:
```suggestion
		FIELD f_lqzypovm spawnsBlockDustParticles Z
```

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -343,6 +348,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 			ARG 3 block
 			ARG 4 posFrom
 			ARG 5 notify
+		METHOD m_hndkckmc spawnsBlockDustParticles ()Z
```
supersaiyansubtlety:
```suggestion
		METHOD m_hndkckmc doesSpawnBlockDustParticles ()Z
```
[CONVENTIONS.md "getters"](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#getters-setters-withers-and-creators)
> For boolean getters, use "is".

"is" is grammatically incorrect here -> "does"

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -593,6 +599,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 		FIELD f_piwzbwkw suffocationPredicate Lnet/minecraft/unmapped/C_triydqro$C_izsedryc;
 		FIELD f_pxqjcqnw lootTableId Lnet/minecraft/unmapped/C_ncpywfca;
 		FIELD f_stojhpmj solid Z
+		FIELD f_vgbjitlh spawnBlockDustParticles Z
```
supersaiyansubtlety:
```suggestion
		FIELD f_vgbjitlh spawnsBlockDustParticles Z
```

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
```diff
@@ -36,6 +36,8 @@ CLASS net/minecraft/unmapped/C_wwadquuj net/minecraft/block/entity/BeaconBlockEn
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+	METHOD m_jlfysuud getEffectOrNull (Lnet/minecraft/unmapped/C_jaqasomh;)Lnet/minecraft/unmapped/C_jaqasomh;
```
supersaiyansubtlety:
```suggestion
	METHOD m_jlfysuud getEffect (Lnet/minecraft/unmapped/C_jaqasomh;)Lnet/minecraft/unmapped/C_jaqasomh;
```
I don't think `OrNull` is needed here, `get(...)` methods typically return null if the thing can't be gotten

mappings/net/minecraft/item/SuspiciousStewEffectContainer.mapping
```diff
@@ -1,4 +1,15 @@
 CLASS net/minecraft/unmapped/C_noldxssx net/minecraft/item/SuspiciousStewEffectContainer
-	METHOD m_gaivnnnr getAll ()Ljava/util/List;
+	METHOD m_gaivnnnr getAllContainers ()Ljava/util/List;
+	METHOD m_tknkyyvb getAllEffects ()Ljava/util/List;
 	METHOD m_zuxvbyei getContainerOrNull (Lnet/minecraft/unmapped/C_gmbqjnle;)Lnet/minecraft/unmapped/C_noldxssx;
```
supersaiyansubtlety:
```suggestion
	METHOD m_zuxvbyei getContainer (Lnet/minecraft/unmapped/C_gmbqjnle;)Lnet/minecraft/unmapped/C_noldxssx;
```
same as other `OrNull`, might as well change this one too if there's agreement that it's not needed

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -620,6 +627,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 			COMMENT <p>This also marks a block as non-opaque.
 		METHOD m_dsszzhch pistonBehavior (Lnet/minecraft/unmapped/C_mojvzhua;)Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
 			ARG 1 pistonBehavior
+		METHOD m_dyoolquj noBlockDustParticles ()Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
```
supersaiyansubtlety:
```suggestion
		METHOD m_dyoolquj withoutBlockDustParticles ()Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
```
[CONVENSION.md "withers"](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#getters-setters-withers-and-creators)
> Use "with" for methods that return a copy of an object with a different value for some property. Name the parameter the same as the property.

supersaiyansubtlety (changes requested):

mappings/net/minecraft/block/AbstractBlock.mapping
EnnuiL: oh, good point

mappings/net/minecraft/block/AbstractBlock.mapping
EnnuiL: oh btw this name that I've filled followed a convention that was already in place on the Settings class; however, I like the idea of `without`! We should investigate further

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
EnnuiL: hm, am not sure about doing that

eristhea:
`EnnuiL:john_madden`

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -241,6 +241,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 		FIELD f_jnxtuzdb hardness F
 		FIELD f_jvcpotqu isAir Z
 		FIELD f_lpessbtt pistonBehavior Lnet/minecraft/unmapped/C_mojvzhua;
+		FIELD f_lqzypovm spawnsBlockDustParticles Z
```
eristhea:
```suggestion
		FIELD f_lqzypovm spawnsDustParticles Z
```
I think the `block` here is very redundant

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -343,6 +348,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 			ARG 3 block
 			ARG 4 posFrom
 			ARG 5 notify
+		METHOD m_hndkckmc doesSpawnBlockDustParticles ()Z
```
eristhea:
```suggestion
		METHOD m_hndkckmc spawnsDustParticles ()Z
```
shorter and matching my suggestion for the field

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -593,6 +599,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 		FIELD f_piwzbwkw suffocationPredicate Lnet/minecraft/unmapped/C_triydqro$C_izsedryc;
 		FIELD f_pxqjcqnw lootTableId Lnet/minecraft/unmapped/C_ncpywfca;
 		FIELD f_stojhpmj solid Z
+		FIELD f_vgbjitlh spawnsBlockDustParticles Z
```
eristhea:
```suggestion
		FIELD f_vgbjitlh spawnsDustParticles Z
```
we're doing this now

mappings/net/minecraft/block/AbstractBlock.mapping
```diff
@@ -620,6 +627,7 @@ CLASS net/minecraft/unmapped/C_triydqro net/minecraft/block/AbstractBlock
 			COMMENT <p>This also marks a block as non-opaque.
 		METHOD m_dsszzhch pistonBehavior (Lnet/minecraft/unmapped/C_mojvzhua;)Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
 			ARG 1 pistonBehavior
+		METHOD m_dyoolquj withoutBlockDustParticles ()Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
```
eristhea:
```suggestion
		METHOD m_dyoolquj withoutDustParticles ()Lnet/minecraft/unmapped/C_triydqro$C_xnkxsdfy;
```
whee

mappings/net/minecraft/block/ChorusFlowerBlock.mapping
```diff
@@ -1,4 +1,5 @@
 CLASS net/minecraft/unmapped/C_knlkybdi net/minecraft/block/ChorusFlowerBlock
+	FIELD f_cviyldrt SUPPORT_SHAPE Lnet/minecraft/unmapped/C_zscvhwbd;
```
eristhea, line 2: is this the moj? I think `STEM_SHAPE` or something would flow nicer but I don't have a strong opinion either way

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
eristhea: I support the change, `getEffectNullable` is better than `orNull` if you're still not a fan

mappings/net/minecraft/block/AbstractBlock.mapping
EnnuiL: good point!

mappings/net/minecraft/block/ChorusFlowerBlock.mapping
EnnuiL: yes but moj is `SUPPORT_SHAPE`; I have went with it because the fence mappings used it; could consider `STEM_SHAPE` though!

eristhea: I like support shape then!

EnnuiL: wait frick i meant to reply with "moj is `BLOCK_SUPPORT_SHAPE`"

eristhea:
@EnnuiL do you intend to fix this up, or should I take it over?

EnnuiL:
oh god i completely forgot about this pr; i have been quite busy recently, but i'll see if i can tackle it later

EnnuiL:
actually, please take it over

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
OroArmor: i think `OrNull` is fine

OroArmor (approved):

eristhea (approved):

supersaiyansubtlety (approved):

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
```diff
@@ -36,6 +36,8 @@ CLASS net/minecraft/unmapped/C_wwadquuj net/minecraft/block/entity/BeaconBlockEn
 		ARG 1 x
 		ARG 2 y
 		ARG 3 z
+	METHOD m_jlfysuud getEffect (Lnet/minecraft/unmapped/C_jaqasomh;)Lnet/minecraft/unmapped/C_jaqasomh;
```
EnnuiL: i still have a bad feeling about getting rid of the `orNull`; it just feels reasonable here, it's like, a standard too;
please reconsider it before merge

supersaiyansubtlety: Can you give an example of a similar method with the `OrNull` suffix?
I've been comparing it to `Map#get`, but there could be a better comparison.

supersaiyansubtlety: Actually maybe part of the issue is `get` might be the wrong verb here.
It checks if the beacon has the effect, returns it if so, or null otherwise.
It's like a weird `contains` method that returns arg/null instead of true/false.

supersaiyansubtlety: This method seems to be replacing what mojmap called `getValidEffectById`, I think valid is important.
I also think `effect` in the name is redundant given the parameter.
How about just `validate`?

eristhea: after looking through the source, I agree with @supersaiyansubtlety on `validate` being the best name
I'm going to add javadoc too

eristhea: @EnnuiL please take a look at the new name!

supersaiyansubtlety (approved):

mappings/net/minecraft/block/entity/BeaconBlockEntity.mapping
EnnuiL: Looks good to me; I'll double-check on Enigma soon, since uh, `getContainerOrNull` still lost the `OrNull` and i remember that it was there for good reasons

eristhea: it's annotated as `@Nullable` which imo makes the suffix redundant always

eristhea: it's annotated as `@Nullable` which imo makes the suffix redundant always

EnnuiL: fair enough

PR #485, opened 2023-08-29
refactor and complete client.search.* mappings

Antikyth:
This PR completes mappings for everything in the `client.search` package.  I have compared against Mojmap and Yarn mappings, but I don't believe that either has names that both accurately and intuitively describe what these things do in this case.

I have also made changes to the existing mappings:
- `IdentifierSearchTree`, `PlainTextSearchTree` --> `IdentifierSearcher`, `PlainTextSearcher`
  - In Mojmap, `Searcher`, `RefreshingSearcher`, and these two are all known as search trees.  Additionally, these two have the features present in `Searcher` and `RefreshingSearcher`, and as such I believe their names should be consistent.
- `Searchable`, `RefreshingSearchable` --> `Searcher`, `RefreshingSearcher`
  - As far as I understand, these interfaces are not implemented on types that can be searched, but rather on types whose purpose is to search another type.  As such, I think types implementing these interfaces could be easily confused with the types that are actually being searched if the names were to remain unchanged.  I do agree that the names `Searcher` and `RefreshingSearcher` are more intuitive than Mojmap's `SearchTree` and `RefreshingSearchTree`.
- `SearchManager.put` --> `SearchManager.registerProvider`
  - I have compared to `SoundManager` in `client.sound`, where the word register is used rather than put.  I think this is more intuitive.


mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
...
 		ARG 1 key
+		ARG 2 provider
 	METHOD m_xgrlmmoi get (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_inuoeylq;
 		ARG 1 key
 	METHOD m_yhcpnstp refreshSearch (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Ljava/util/List;)V
 		ARG 1 key
 		ARG 2 searchables
+	CLASS C_faknnetb RefreshingSearcherProviderFunction
 	CLASS C_gvofkqkn Key
+	CLASS C_lpzuavsm RefreshingSearcherProvider
```
Antikyth, line 18: Holds a `RefreshingSearcher`.  Constructor takes a `RefreshingSearcherProviderFunction` which is used to set the contained `RefreshingSearcher` to a new one based on the given `searchables` (the thing being searched, e.g. itemstacks).

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
...
-	METHOD m_jrasqirm put (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
+	METHOD m_jrasqirm registerProvider (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
 		ARG 1 key
+		ARG 2 provider
 	METHOD m_xgrlmmoi get (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_inuoeylq;
 		ARG 1 key
 	METHOD m_yhcpnstp refreshSearch (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Ljava/util/List;)V
 		ARG 1 key
 		ARG 2 searchables
+	CLASS C_faknnetb RefreshingSearcherProviderFunction
```
Antikyth:
```java
@Environment(EnvType.CLIENT)
public interface RefreshingSearcherProviderFunction<T> extends Function<List<T>, RefreshingSearcher<T> {
}
```

mappings/net/minecraft/client/search/SearchManager.mapping
Antikyth:
```java
@Environment(EnvType.CLIENT)
static class RefreshingSearcherProvider<T> {
    private final SearchManager.RefreshingSearcherProviderFunction<T> provider;
    RefreshingSearcher<T> searcher = RefreshingSearcher.empty();
    
    RefreshingSearcherProvider(SearchManager.RefreshingSearcherProviderFunction<T> provider) {
        this.provider = provider;
    }
    
    void provideSearcher(List<T> searchables) {
        this.searcher = this.provider.apply(searchables);
        this.searcher.refresh();
    }
    
    void refreshSearcher() {
        this.searcher.refresh();
    }
}
```

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
 	FIELD f_exfajduf RECIPE_OUTPUT Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_hpwjudey ITEM_TOOLTIP Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_jivcozuv ITEM_TAG Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
-	FIELD f_xowozfgq instances Ljava/util/Map;
-	METHOD m_jmupjdnd (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_zqzigdge$C_lpzuavsm;
+	FIELD f_xowozfgq searcherProviders Ljava/util/Map;
+	METHOD m_jmupjdnd getProvider (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_zqzigdge$C_lpzuavsm;
 		ARG 1 key
-	METHOD m_jrasqirm put (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
+	METHOD m_jrasqirm registerProvider (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
```
Antikyth, line 8:
```java
public <T> void registerProvider(SearchManager.Key<T> key, RefreshingSearcherProviderFunction<T> provider) {
    this.searcherProviders.put(key, new SearchManager.RefreshingSearcherProvider<>(provider);
}
```

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
 	FIELD f_exfajduf RECIPE_OUTPUT Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_hpwjudey ITEM_TOOLTIP Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_jivcozuv ITEM_TAG Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
-	FIELD f_xowozfgq instances Ljava/util/Map;
-	METHOD m_jmupjdnd (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_zqzigdge$C_lpzuavsm;
+	FIELD f_xowozfgq searcherProviders Ljava/util/Map;
+	METHOD m_jmupjdnd getProvider (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_zqzigdge$C_lpzuavsm;
```
Antikyth, line 6:
```java
private <T> SearchManager.RefreshingSearcherProvider<T> getProvider(SearchManager.Key<T> key) {
    SearchManager.RefreshingSearcherProvider<T> provider = this.searcherProviders.get(key);
    
    if (provider == null) {
        throw new IllegalStateException("Tree builder not registered");
    } else {
        return provider;
    }
}
```

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
 	FIELD f_exfajduf RECIPE_OUTPUT Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_hpwjudey ITEM_TOOLTIP Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
 	FIELD f_jivcozuv ITEM_TAG Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;
-	FIELD f_xowozfgq instances Ljava/util/Map;
-	METHOD m_jmupjdnd (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_zqzigdge$C_lpzuavsm;
+	FIELD f_xowozfgq searcherProviders Ljava/util/Map;
```
Antikyth, line 5:
```java
private final Map<SearchManager.Key<?>, SearchManager.RefreshingSearcherProvider<?>> searcherProviders = new HashMap();
```

mappings/net/minecraft/client/search/IdentifierSearch.mapping
```diff
@@ -1,9 +1,9 @@
 CLASS net/minecraft/unmapped/C_ttylwehh net/minecraft/client/search/IdentifierSearch
-	FIELD f_ducunurc searchTree Lnet/minecraft/unmapped/C_mkylkzob;
+	FIELD f_ducunurc identifierSearcher Lnet/minecraft/unmapped/C_mkylkzob;
 	FIELD f_pjexfxok comparator Ljava/util/Comparator;
 	METHOD <init> (Ljava/util/function/Function;Ljava/util/List;)V
 		ARG 1 identifiers
-		ARG 2 values
+		ARG 2 searchables
```
supersaiyansubtlety: This is not a list of things, each of which is searchable; it's a single list (which is seearchable), so it shouldn't be plural.
Calling it `searchable` would technically make sense, but I prefer the existing `values` name (which also gets to follow the convention of making collections plural).

mappings/net/minecraft/client/search/IdentifierSearcher.mapping
```diff
@@ -7,5 +7,5 @@ CLASS net/minecraft/unmapped/C_mkylkzob net/minecraft/client/search/IdentifierSe
 		ARG 1 path
 	METHOD m_mbaochwd empty ()Lnet/minecraft/unmapped/C_mkylkzob;
 	METHOD m_nzgsulfj create (Ljava/util/List;Ljava/util/function/Function;)Lnet/minecraft/unmapped/C_mkylkzob;
-		ARG 0 list
+		ARG 0 searchables
```
supersaiyansubtlety:
```suggestion
		ARG 0 values
```
Same as the other `searchables`.

mappings/net/minecraft/client/search/PlainTextSearch.mapping
```diff
@@ -1,8 +1,8 @@
 CLASS net/minecraft/unmapped/C_vsqjjavd net/minecraft/client/search/PlainTextSearch
-	FIELD f_betspott plainTextSearchTree Lnet/minecraft/unmapped/C_kajwzxdj;
+	FIELD f_betspott plainTextSearcher Lnet/minecraft/unmapped/C_kajwzxdj;
 	FIELD f_kugqkwbk filler Ljava/util/function/Function;
-	FIELD f_orttexni values Ljava/util/List;
+	FIELD f_orttexni searchables Ljava/util/List;
```
supersaiyansubtlety: Same as the other `searchables`.

mappings/net/minecraft/client/search/PlainTextSearch.mapping
```diff
@@ -1,8 +1,8 @@
...
 	METHOD <init> (Ljava/util/function/Function;Ljava/util/function/Function;Ljava/util/List;)V
 		ARG 1 filler
-		ARG 2 ids
-		ARG 3 values
+		ARG 2 identifiers
+		ARG 3 searchables
```
supersaiyansubtlety: Same as the other `searchables`.

mappings/net/minecraft/client/search/IdentifierSearch.mapping
```diff
@@ -1,9 +1,9 @@
 CLASS net/minecraft/unmapped/C_ttylwehh net/minecraft/client/search/IdentifierSearch
-	FIELD f_ducunurc searchTree Lnet/minecraft/unmapped/C_mkylkzob;
+	FIELD f_ducunurc identifierSearcher Lnet/minecraft/unmapped/C_mkylkzob;
 	FIELD f_pjexfxok comparator Ljava/util/Comparator;
 	METHOD <init> (Ljava/util/function/Function;Ljava/util/List;)V
 		ARG 1 identifiers
```
supersaiyansubtlety:
```suggestion
		ARG 1 ids
```
`ids` is a [conventional abbreviation](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#abbreviations) for `identifiers`

mappings/net/minecraft/client/search/IdentifierSearcher.mapping
```diff
@@ -7,5 +7,5 @@ CLASS net/minecraft/unmapped/C_mkylkzob net/minecraft/client/search/IdentifierSe
...
 		ARG 1 identifiers
```
supersaiyansubtlety:
```suggestion
		ARG 1 ids
```
`ids` is a [conventional abbreviation](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#abbreviations) for `identifiers`

mappings/net/minecraft/client/search/PlainTextSearch.mapping
```diff
@@ -1,8 +1,8 @@
...
-	FIELD f_betspott plainTextSearchTree Lnet/minecraft/unmapped/C_kajwzxdj;
+	FIELD f_betspott plainTextSearcher Lnet/minecraft/unmapped/C_kajwzxdj;
 	FIELD f_kugqkwbk filler Ljava/util/function/Function;
-	FIELD f_orttexni values Ljava/util/List;
+	FIELD f_orttexni searchables Ljava/util/List;
 	METHOD <init> (Ljava/util/function/Function;Ljava/util/function/Function;Ljava/util/List;)V
 		ARG 1 filler
-		ARG 2 ids
-		ARG 3 values
+		ARG 2 identifiers
```
supersaiyansubtlety: `ids` is a [conventional abbreviation](https://github.com/QuiltMC/quilt-mappings/blob/1.20-pre6/CONVENTIONS.md#abbreviations) for `identifiers`

mappings/net/minecraft/client/search/PlainTextSearcher.mapping
```diff
@@ -1,11 +1,11 @@
-CLASS net/minecraft/unmapped/C_kajwzxdj net/minecraft/client/search/PlainTextSearchTree
+CLASS net/minecraft/unmapped/C_kajwzxdj net/minecraft/client/search/PlainTextSearcher
 	METHOD m_lqmnbwjj (Lnet/minecraft/unmapped/C_eyvsctee;Ljava/lang/Object;Ljava/lang/String;)V
 		ARG 2 s
 	METHOD m_rzntbizi (Ljava/lang/String;)Ljava/util/List;
 		ARG 0 s
 	METHOD m_tsssxwiu create (Ljava/util/List;Ljava/util/function/Function;)Lnet/minecraft/unmapped/C_kajwzxdj;
-		ARG 0 list
-		ARG 1 function
+		ARG 0 searchables
```
supersaiyansubtlety:
```suggestion
		ARG 0 values
```
same as other `searchables`

supersaiyansubtlety (changes requested):
I think `IdentifierSearch` should be renamed.
Using `Search` would be ok if class instances were single-use, so it truly would be one search, but this class appears to be reusable.
It seems like the only reason not to call it `IdentifierSearcher` is that that name is currently taken.

Something along the lines of
`IdentifierSearch` -> `IdentifierSearcher`
`IdentifierSearcher` -> `AbstractIdentifierSearcher`
would be better I think, but using `Abstract` for an interface is weird.

If this gets renamed, `PlainTextSearch` would also have to be renamed.

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
...
 		ARG 1 key
-	METHOD m_jrasqirm put (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
+	METHOD m_jrasqirm registerProvider (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;)V
 		ARG 1 key
+		ARG 2 provider
 	METHOD m_xgrlmmoi get (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;)Lnet/minecraft/unmapped/C_inuoeylq;
 		ARG 1 key
 	METHOD m_yhcpnstp refreshSearch (Lnet/minecraft/unmapped/C_zqzigdge$C_gvofkqkn;Ljava/util/List;)V
 		ARG 1 key
 		ARG 2 searchables
```
supersaiyansubtlety:
```suggestion
		ARG 2 values
```
Same as other `searchables`.

mappings/net/minecraft/client/search/SearchManager.mapping
```diff
@@ -2,14 +2,22 @@ CLASS net/minecraft/unmapped/C_zqzigdge net/minecraft/client/search/SearchManage
...
+	CLASS C_faknnetb RefreshingSearcherProviderFunction
 	CLASS C_gvofkqkn Key
+	CLASS C_lpzuavsm RefreshingSearcherProvider
+		FIELD f_qavvoydc searcher Lnet/minecraft/unmapped/C_wvjmuhfo;
+		FIELD f_qbsdiplx provider Lnet/minecraft/unmapped/C_zqzigdge$C_faknnetb;
+		METHOD m_chwfndau provideSearcher (Ljava/util/List;)V
+			ARG 1 searchables
```
supersaiyansubtlety:
```suggestion
			ARG 1 values
```
Same as other `searchables`

supersaiyansubtlety (changes requested):
I missed a few `searchables`

OroArmor (approved):
Once the suggestions by supersaiyansubtlety are addressed, this PR looks great! Thanks for the additional comments on the PR, that really helps with the approval process!

mappings/net/minecraft/client/search/IdentifierSearch.mapping
Antikyth: I think you make a good point in that if it is confusing then a better name should be used, like values.

Just to describe how I was thinking about it though: each element in the list maps with a function to a stream of strings I believe.  So like if the searchables are ItemStacks, then each stack is searchable in that its name/lines of tooltip/whatever it is, I don't actually know if that's exactly what it represents in the case of ItemStacks or not.  So I was thinking both that the item stacks are searched but also that each ItemStack has its corresponding strings searched.

(Note: replying when tired, will review properly when it isn't so late)

mappings/net/minecraft/client/search/IdentifierSearch.mapping
Antikyth: Ah.  Saw that id was, but I didn't see ids and thought that if it wasn't an established convention that people are used to, it could be confusing as to whether the s stood for something or not.  Not when you think about it, just at first glance.  If the plural form is commonly abbreviated, then I can switch it to that.

Antikyth:
> I think `IdentifierSearch` should be renamed.
> Using `Search` would be ok if class instances were single-use, so it truly would be one search, but this class appears to be reusable.
> It seems like the only reason not to call it `IdentifierSearcher` is that that name is currently taken.
> 
> Something along the lines of
> `IdentifierSearch` -> `IdentifierSearcher`
> `IdentifierSearcher` -> `AbstractIdentifierSearcher`
> would be better I think, but using `Abstract` for an interface is weird.
> 
> If this gets renamed, `PlainTextSearch` would also have to be renamed.

Yeah, I knew that Search wasn't right but I knew they had to be different names and it was an existing name.  How about `IdentifierSearcherInterface`?  I know it isn't recommended to label interfaces in the names, but I would argue that this is a case where making an exception would be for the best.  It's the most easily understandable name given that they can't be the same.

Antikyth:
> Once the suggestions by supersaiyansubtlety are addressed, this PR looks great! Thanks for the additional comments on the PR, that really helps with the approval process!

Heh, you're welcome!  When I made the PR I thought to myself "man, I would hate to review this without seeing the context", so I thought I'd better add comments to make it easier 😅

Antikyth:
I believe all suggestions have been applied.  I have renamed `IdentifierSearcher` and `PlainTextSearcher` to `IdentifierSearcherInterface` and `PlainTextSearcherInterface` for now, as `IdentifierSearch` and `PlainTextSearch` are being renamed to `IdentifierSearcher` and `PlainTextSearcher`.  Let me know if they aren't suitable names.

supersaiyansubtlety (approved):
Agreed, comments from the pull requester are definitely helpful!

supersaiyansubtlety:
> Yeah, I knew that Search wasn't right but I knew they had to be different names and it was an existing name. How about `IdentifierSearcherInterface`? I know it isn't recommended to label interfaces in the names, but I would argue that this is a case where making an exception would be for the best. It's the most easily understandable name given that they can't be the same.

Yeah the `Interface` suffix is also a bit strange, but I can't think of anything better. It's a sort of strange class separation in the first place so the names likely have to be a little strange to match.

mappings/net/minecraft/client/search/SearchManager.mapping
eristhea: this PR looks fantastic, thank you! my only comment: what's stopping you from calling this `Factory` instead of `ProviderFunction`? I think that would be better for conciseness

eristhea (changes requested):

Antikyth:
> this PR looks fantastic, thank you! my only comment: what's stopping you from calling this `Factory` instead of `ProviderFunction`? I think that would be better for conciseness

I'm just not that familiar with the factory pattern in Java.  I looked it up, but it didn't seem like a regular `Function` was really a factory (may be wrong).  It just extends `Function`.

supersaiyansubtlety:
> I'm just not that familiar with the factory pattern in Java. I looked it up, but it didn't seem like a regular `Function` was really a factory (may be wrong). It just extends `Function`.

A 'factory' is just static method that always returns a `new` instance of the class it is a factory for.
It looks like all calls to `SearchManager#registerProvider` are passed lambdas that return `new` instances, so `Factory` seems appropriate here.

Antikyth:
Renamed to factory.

eristhea (approved):

eristhea:
moved to FCP for two days!

Antikyth:
Just checking in, was this meant to be merged after two days of the FCP?

eristhea:
my mistake :p

PR #506, opened 2023-10-04
Fix sequence fields

### LINKED ISSUE #470: int sequence fields are not sequences
discussed here: https://github.com/QuiltMC/quilt-mappings/pull/469#discussion_r1275701457

Some relevant fields:

`ServerPlayNetworkHandler#blockChangeSequence`
`PlayerInteractionWithItemC2SPacket#sequence`
`PlayerActionC2SPacket#sequence`
`PlayerInteractionWithBlockC2SPacket#sequence`
`BlockUpdateAcknowledgmentS2CPacket#sequence`

These integers aren't sequences. AFAICT, they indicate their instance's order within a sequence.

What I found:
These `sequence` fields either default to 0 or their value traces back to `C_czisrdmd#field_37954`.
This field is incremented in `C_czisrdmd#method_41937`, which is only called by `ClientPlayerInteractionManager#method_41931`.
`method_41931` is passed packet factories that take an `int` (`C_czisrdmd#field_37954`) and create the above packets with that sequence.

I think `sequenceIndex` would be a better name.

supersaiyansubtlety:
fixes #470

I've also:
1. eliminated `client.multiplayer.prediction` and moved its contents to `cleint.network`: there are only two classes, one of which is a functional interface, and both of which are used for networking
2. renamed some methods to consistently use `BlockUpdate` instead of variations on it
3. renamed `PredictiveAction` -> `PlayerActionPacketFactory`: 
    - `PredictiveAction`->`PlayerAction`: the action itself isn't making a prediction, and the actual packets created are all `PlayerAction...` or `PlayerInteraction...`
    - `PacketFactory`: it's a functional interface whose method creates a packet

Some notes:
all of the indexed actions are player actions that
- can potentially cause block updates (but some usually don't, item interaction is included)
- can be rejected by the server

I've chosen "index of player action attempt that could change block state" -> `actionIndex`
- no "player" because the fields are in `PlayerAction...` and `PlayerInteraction...` packet classes
- no "attempt" because under normal circumstances the actions succeed (failure usually means lag or cheating)
- no "could change block state" because I don't know of a concise and non-confusing way to include it (any mention of block makes it sound like item interaction isn't included)


eristhea (approved):
looks great!

IotaBread (approved):

OroArmor (approved):

PR #508, opened 2023-10-05
Fix some misnomers

### LINKED ISSUE #504: `Entity`'s `getCameraPosVec` and `getClientCameraPosVec` are misnomers
The first should be `getLerpedEyePos` (matches `getEyePos` and `getLerpedPos`).

For the second, the only time it doesn't delegate to the first is in `ArmorStandEntity` when `isMarker()` is true, in which case it returns the pos its bounding box intersects with the highest light level; and it's only called from `EntityRenderer#getLight(...)`.

So maybe something like `getObservingLightPos` (mojmap is `getLightProbePosition`)?

### LINKED ISSUE #505: `ServerWorld`'s `getDragonPart` is a misnomer
The method gets an entity by id first and only accesses `dragonParts` if that is null.
It's also called in `PlayerInteractionWithEntityC2SPacket#getEntity`.

I think `getEntityOrDragonPart` would be better (but a bit clunky) (mojmap is `getEntityOrPart`).

supersaiyansubtlety:
fixes #504 and #505
explanations in issues


eristhea (approved):
as long as the eye pos is always the eye pos, it doesn't change when you go into `f5`

supersaiyansubtlety:
> as long as the eye pos is always the eye pos, it doesn't change when you go into `f5`

double checked and it doesn't

IotaBread (approved):

PR #509, opened 2023-10-08
Added getAirSpeedMultiplier mapping in AquaticMoveControl

HyperPigeon:
Added getAirSpeedMultiplier mapping for private static method in AquaticMoveControl.


OroArmor (approved):

mappings/net/minecraft/entity/ai/control/AquaticMoveControl.mapping
```diff
@@ -11,3 +11,4 @@ CLASS net/minecraft/unmapped/C_rtyfovet net/minecraft/entity/ai/control/AquaticM
 		ARG 4 speedInWater
 		ARG 5 speedInAir
 		ARG 6 buoyant
+	METHOD m_rvgifltm getAirSpeedMultiplier (F)F
```
supersaiyansubtlety:
```suggestion
	METHOD m_rvgifltm calcAirSpeedMultiplier (F)F
```
I don't think we use `get` for either static methods or methods that take params. The only exceptions I've seen are for map-like access (`getEntityByRawId`).

supersaiyansubtlety (changes requested):

supersaiyansubtlety (approved):

eristhea (approved):

PR #514, opened 2023-10-31
some item things

eristhea:
gets `ItemStack` and `Item` up to speed with the crafter existing


mappings/net/minecraft/item/Item.mapping
```diff
@@ -63,8 +65,10 @@ CLASS net/minecraft/unmapped/C_vorddnax net/minecraft/item/Item
 		ARG 1 nbt
 	METHOD m_fofdychq hasRecipeRemainder ()Z
 		COMMENT Checks if this item has a remainder item that is left behind when used as a crafting ingredient.
-	METHOD m_fwxnsuhn onCraft (Lnet/minecraft/unmapped/C_sddaxwyk;Lnet/minecraft/unmapped/C_cdctfzbn;Lnet/minecraft/unmapped/C_jzrpycqo;)V
+	METHOD m_fwxnsuhn onCraftByPlayer (Lnet/minecraft/unmapped/C_sddaxwyk;Lnet/minecraft/unmapped/C_cdctfzbn;Lnet/minecraft/unmapped/C_jzrpycqo;)V
 		COMMENT Called when a player acquires the item by crafting, smelting, smithing, etc.
+		COMMENT
+		COMMENT Not triggered when the item is automatically crafted by a block such as the crafter.
```
supersaiyansubtlety: This javadoc confuses me when it comes to smelters. It says it's called for smelters, which I would consider "automatic", but it also says that it's not called for automatic things.

supersaiyansubtlety (changes requested):

mappings/net/minecraft/item/Item.mapping
eristhea: hm, on second thought you're right and this isn't really a clarification --
for example the furnace: it runs in the background, but this is called when the player takes the items out to give experience

so the crafter isn't really a new concept, and we can just leave the docs as-is

supersaiyansubtlety (approved):

OroArmor (approved):

PR #516, opened 2023-12-08
first looks into 1.20.4

moehreag:
These are just mostly the unmapped classes I quickly found names for.
- mostly new classes that are mostly related to new stuff

Things I'm not sure about:
- `BlockCodecs` should probably not be in the `block` package. Where should it go?

### DIFF
--- a/mappings/net/minecraft/util/JavaOps.mapping
+++ b/mappings/net/minecraft/util/JavaOps.mapping
@@ -1,0 +1,103 @@
+CLASS net/minecraft/unmapped/C_diganwjh net/minecraft/util/JavaOps
+	FIELD f_mtjkvncv INSTANCE Lnet/minecraft/unmapped/C_diganwjh;
+	METHOD convertTo convertTo (Lcom/mojang/serialization/DynamicOps;Ljava/lang/Object;)Ljava/lang/Object;
+		ARG 2 o
+	METHOD createBoolean createBoolean (Z)Ljava/lang/Object;
+		ARG 1 bl
+	METHOD createByte createByte (B)Ljava/lang/Object;
+		ARG 1 b
+	METHOD createByteList createByteList (Ljava/nio/ByteBuffer;)Ljava/lang/Object;
+		ARG 1 buffer
+	METHOD createDouble createDouble (D)Ljava/lang/Object;
+		ARG 1 d
+	METHOD createFloat createFloat (F)Ljava/lang/Object;
+		ARG 1 f
+	METHOD createInt createInt (I)Ljava/lang/Object;
+		ARG 1 i
+	METHOD createIntList createIntList (Ljava/util/stream/IntStream;)Ljava/lang/Object;
+		ARG 1 stream
+	METHOD createList createList (Ljava/util/stream/Stream;)Ljava/lang/Object;
+		ARG 1 stream
+	METHOD createLong createLong (J)Ljava/lang/Object;
+		ARG 1 l
+	METHOD createLongList createLongList (Ljava/util/stream/LongStream;)Ljava/lang/Object;
+		ARG 1 stream
+	METHOD createMap createMap (Ljava/util/Map;)Ljava/lang/Object;
+		ARG 1 map
+	METHOD createMap createMap (Ljava/util/stream/Stream;)Ljava/lang/Object;
+		ARG 1 stream
+	METHOD createNumeric createNumeric (Ljava/lang/Number;)Ljava/lang/Object;
+		ARG 1 n
+	METHOD createShort createShort (S)Ljava/lang/Object;
+		ARG 1 s
+	METHOD createString createString (Ljava/lang/String;)Ljava/lang/Object;
+		ARG 1 string
+	METHOD empty empty ()Ljava/lang/Object;
+	METHOD emptyList emptyList ()Ljava/lang/Object;
+	METHOD emptyMap emptyMap ()Ljava/lang/Object;
+	METHOD getBooleanValue getBooleanValue (Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
+	METHOD getByteBuffer getByteBuffer (Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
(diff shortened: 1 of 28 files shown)


eristhea:
codec stuff goes in `util`

Patbox:
BlockCodecs isn't a correct name, as the bound registry uses id of `block_type` and that class populates it

moehreag:
> BlockCodecs isn't a correct name, as the bound regsistry uses id of `block_type` and that class populates it

Hmm... looking at the class again maybe `BlockTypeRegistry` would be a better name?

OroArmor:
`BlockTypes`, follows the `Blocks` naming scheme

Patbox:
It's not a registry either, the closest thing would be Blocks or Items class

OroArmor:
Also being in the `n.m.block` is the right package

moehreag:
> Also being in the `n.m.block` is the right package

alright!

mappings/net/minecraft/util/JavaOps.mapping
supersaiyansubtlety, line 1: AFAICT QMap doesn't use `o` for objects unless it's for `equals(Ljava/lang/Object;)Z`, and I'm not sure about any of the single-letter names of these params (except `i`).
I think it could make sense to use them for primitives (and maybe `Object`s), but the convention should be made clear if so.

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+	METHOD m_afyiirgn (Lcom/google/common/collect/ImmutableMap$Builder;Lcom/mojang/datafixers/util/Pair;)V
+		ARG 1 pair
+	METHOD m_boxzzpcu (Ljava/util/Map$Entry;)Lcom/mojang/datafixers/util/Pair;
+		ARG 0 entry
+	METHOD m_mrnxttti entryStream (Ljava/util/Map;)Ljava/util/stream/Stream;
+		ARG 0 map
+	METHOD m_rxpnwroz toMap (Lcom/mojang/serialization/MapLike;)Ljava/util/Map;
+		ARG 0 mapLike
+	METHOD m_sfvskaol (Ljava/lang/Object;)Ljava/lang/Object;
+		ARG 0 object
```
supersaiyansubtlety, line 69: inconsistent with other `o` params

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+	METHOD mapBuilder mapBuilder ()Lcom/mojang/serialization/RecordBuilder;
+	METHOD mergeToList mergeToList (Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
```
supersaiyansubtlety, line 72:
```suggestion
		ARG 1 o1
```

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+		ARG 2 o2
+	METHOD mergeToList mergeToList (Ljava/lang/Object;Ljava/util/List;)Lcom/mojang/serialization/DataResult;
+		ARG 1 object
```
supersaiyansubtlety, line 75: inconsistent

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+		ARG 2 list
+	METHOD mergeToMap mergeToMap (Ljava/lang/Object;Lcom/mojang/serialization/MapLike;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
+		ARG 2 mapLike
+	METHOD mergeToMap mergeToMap (Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
```
supersaiyansubtlety, line 81:
```suggestion
		ARG 1 o1
```

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+		ARG 2 o2
+		ARG 3 o3
+	METHOD mergeToMap mergeToMap (Ljava/lang/Object;Ljava/util/Map;)Lcom/mojang/serialization/DataResult;
+		ARG 1 o
+		ARG 2 map
+	METHOD remove remove (Ljava/lang/Object;Ljava/lang/String;)Ljava/lang/Object;
+		ARG 1 map
+		ARG 2 s
```
supersaiyansubtlety, line 89:
```suggestion
		ARG 2 string
```
`s` is used for short in other places, and I think if single letters are used, they should be for primitives (and maybe `Object`s)

mappings/net/minecraft/util/JavaOps.mapping
```diff
@@ -0,0 +1,103 @@
...
+	CLASS C_bjqonldn
+		METHOD entries entries ()Ljava/util/stream/Stream;
+		METHOD get get (Ljava/lang/Object;)Ljava/lang/Object;
+			ARG 1 o
+		METHOD get get (Ljava/lang/String;)Ljava/lang/Object;
+			ARG 1 string
+	CLASS C_wogpegfy MapBuilder
+		METHOD append (Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
+			ARG 1 o
```
supersaiyansubtlety, line 98:
```suggestion
			ARG 1 o1
```

supersaiyansubtlety (changes requested):

mappings/net/minecraft/util/JavaOps.mapping
moehreag: > AFAICT QMap doesn't use `o` for objects unless it's for `equals(Ljava/lang/Object;)Z`, and I'm not sure about any of the single-letter names of these params (except `i`). I think it could make sense to use them for primitives (and maybe `Object`s), but the convention should be made clear if so.

It is entirely possible that using `o` for object parameters is something I entirely made up and isn't used anywhere and mentioned in any conventions (except for `equals` methods)

PR #519, opened 2023-12-22
Map LootableInventory

### LINKED ISSUE #518: `C_ehawoasn.method_10997` is not implemented by `BlockEntity.getWorld`
# The Issue
This issue was encountered on 1.20.4, when breaking a chest. The chest attempts to dump its loot on the ground, causing `C_ehawoasn.method_10997` to be called, which should be implemented by `BlockEntity.getWorld`, but it is not. This causes the game to crash with an AbstractMethodError.

# The Fix
A simple fix would be to add a mapping for `C_ehawoasn` and change its `method_10997` method to `getWorld`. Note: `C_ehawoasn` also has a `method_11016` which should probably be `getPos`.

Kneelawk:
# This PR
This PR adds mappings for `C_ehawoasn` to fix a critical bug that crashed the game when breaking chests.

Most importantly, this maps `C_ehawoasn.method_10997` to `LootableInventory.getWorld` and `C_ehawoasn.method_11016` to `LootableInventory.getPos`.

# Associated
* This fixes #518.

### DIFF
--- a/mappings/net/minecraft/inventory/LootableInventory.mapping
+++ b/mappings/net/minecraft/inventory/LootableInventory.mapping
@@ -1,0 +1,14 @@
+CLASS net/minecraft/unmapped/C_ehawoasn net/minecraft/inventory/LootableInventory
+	FIELD f_yeibkfzd LOOT_TABLE_SEED_KEY Ljava/lang/String;
+	FIELD f_ztchflsp LOOT_TABLE_KEY Ljava/lang/String;
+	METHOD m_canflxyp setLootTableId (Lnet/minecraft/unmapped/C_ncpywfca;)V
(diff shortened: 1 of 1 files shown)


moehreag (approved):
looks good to me!

IotaBread (approved):

eristhea (approved):

Kneelawk:
These changes should probably be propagated to the `23w51b` branch. Should be easy as a cherry-pick. It might be a little more involved to see how far back these changes need to be brought to.

Kneelawk:
Ok, I did some looking, and it appears this issue was introduced in 1.20.3/1.20.4. So it doesn't look like any backports are necessary.

PR #520, opened 2023-12-31
Status effect mappings

IotaBread:
100% on `n.m.entity.effect`! `StatusEffectInstance.BlendState` is a mojmap name, since I had no clue what the class did. From what I could gather, it's used to calculate a "blending factor" to give smooth fading out to effects. In `23w51b` its only used for the darkness effect


mappings/net/minecraft/entity/effect/InstantHealthOrDamageStatusEffect.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/unmapped/C_cbhpjzoa net/minecraft/entity/effect/InstantHealthOrDamageStatusEffect
+	FIELD f_efjpyqqw isHarm Z
```
supersaiyansubtlety:
```suggestion
	FIELD f_efjpyqqw isDamage Z
```
seems like harm/damage should match
I slightly prefer damage

eristhea (approved):
with @supersaiyansubtlety's suggestion this looks great!

OroArmor (approved):

supersaiyansubtlety (approved):

PR #521, opened 2023-12-31
advancement mappings + codecs!

eristhea:
100%s `Codecs` and `n.m.advancement`. one major refactor: `AdvancementComponent` -> `AdvancementType`. The old name was plain wrong as this class is used to determine which frame should be displayed around the advancement, out of Minecraft's three options

### DIFF
--- a/mappings/net/minecraft/util/dynamic/Codecs.mapping
+++ b/mappings/net/minecraft/util/dynamic/Codecs.mapping
@@ -4,7 +4,7 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
 	COMMENT <p>It has a few methods to create checkers for {@code Codec.flatXmap} to add
 	COMMENT extra value validation to encoding and decoding. See the implementation of
 	COMMENT {@link #withNonEmptyList(Codec)}.
-	FIELD f_aylcbfcj IDENTIFIER_PATH Lcom/mojang/serialization/Codec;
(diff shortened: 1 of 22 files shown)


mappings/net/minecraft/util/dynamic/Codecs.mapping
```diff
@@ -18,25 +20,93 @@ CLASS net/minecraft/unmapped/C_tyjpezxh net/minecraft/util/dynamic/Codecs
...
+	METHOD m_htlmmuvh (Ljava/lang/Object;)Ljava/lang/Object;
+		ARG 0 object
+	METHOD m_ifdvuwzy formattedTime (Ljava/time/format/DateTimeFormatter;)Lcom/mojang/serialization/Codec;
+		ARG 0 formatter
+	METHOD m_ihmqisju (FFLjava/util/function/Function;Ljava/lang/Float;)Lcom/mojang/serialization/DataResult;
+		ARG 3 f
 	METHOD m_ihvqsgya retrieveContext (Ljava/util/function/Function;)Lcom/mojang/serialization/MapCodec;
 		ARG 0 retriever
+	METHOD m_iuewvnnu ([B)Ljava/lang/String;
+		ARG 0 bs
```
supersaiyansubtlety:
```suggestion
		ARG 0 bytes
```
this looked like B.S. more than the plural of `b` to me, I think `bytes` is clearer

supersaiyansubtlety (approved):

OroArmor (approved):

PR #529, opened 2024-01-11
Random mappings here and there

IotaBread:
I tried to get 100% on `n.m.text`, couldn't map a lot on `HoverEvent`. I *think* I got `n.m.test` to 100% or close to it. Also mapped a lot of `n.m.server.command`, though I skipped some of the more complex stuff


supersaiyansubtlety (approved):

mappings/net/minecraft/entity/ItemEntity.mapping
```diff
@@ -1,13 +1,14 @@
 CLASS net/minecraft/unmapped/C_uqpzijng net/minecraft/entity/ItemEntity
 	FIELD f_cejzztdx pickupDelay I
 	FIELD f_davsxlkm CANNOT_PICK_UP_DELAY I
-	FIELD f_laqmyrmj thrower Ljava/util/UUID;
+	FIELD f_laqmyrmj throwerUuid Ljava/util/UUID;
```
eristhea, line 4: I don't see the point of this change

eristhea (approved):

mappings/net/minecraft/entity/ItemEntity.mapping
supersaiyansubtlety: there's another `thrower` method that returns the entity

PR #536, opened 2024-01-18
some mostly network related mappings

moehreag:
some notes:
- the changes in packet methods (e.g. `getType`→`getRenderType`) are to accommodate for the general `getType` method in the `Packet` class (which gets the PacketType of this packet)
- some names of packets in `..Types` classes in `n.m.n.packets` don't make sense together with their string id
- is the creation of a new `n.m.network.phases` package fine?
- there are still a lot of unmapped fields in the `n.m.network.phases` package. These should probably be tackled next.

### DIFF
--- a/mappings/net/minecraft/client/font/FreeTypeUtil.mapping
+++ b/mappings/net/minecraft/client/font/FreeTypeUtil.mapping
@@ -1,0 +1,9 @@
+CLASS net/minecraft/unmapped/C_mwguctid net/minecraft/client/font/FreeTypeUtil
+	FIELD f_rgxcltns HANDLE J
+	METHOD m_bajearil ensureLoaded (ILjava/lang/String;)V
+		ARG 0 result
+		ARG 1 message
+	METHOD m_eelkjuel destroy ()V
+	METHOD m_tczwarzj formatError (I)Ljava/lang/String;
+		ARG 0 code
+	METHOD m_zfxfnpup initialize ()J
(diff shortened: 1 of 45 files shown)


supersaiyansubtlety (commented):
The phases package seems good.
Using `S2C` for class names but `S2c` for method names seems wrong; I think it should always be `S2C` (or `C2S`) and this should be considered an exception to the convention about acronyms.
> some names of packets in ..Types classes in n.m.n.packets don't make sense together with their string id

I'm not sure what you mean by this, are the ids wrong/misleading?

OroArmor:
Hey! while we are working on networking mappings, `ServerConfigurationPacketHandler` is misnamed and misplaced. it should be called `ServerConfiguationNetworkHandler` and be moved to `net.minecraft.server.network`

IotaBread (commented):
About the `S2C/C2S` casing in method names, I agree that it should be always uppercase. Aside from that, the pr looks great

moehreag:
> The phases package seems good. Using `S2C` for class names but `S2c` for method names seems wrong; I think it should always be `S2C` (or `C2S`) and this should be considered an exception to the convention about acronyms.
> 
> > some names of packets in ..Types classes in n.m.n.packets don't make sense together with their string id
> 
> I'm not sure what you mean by this, are the ids wrong/misleading?

here's an example: Image of a line with the contradicting information of PacketType<HandshakeC2SPacket> having the id "intention"

moehreag:
Should also the `phases` package be renamed to `phase` to better fit the conventions?

supersaiyansubtlety:
> Should also the `phases` package be renamed to `phase` to better fit the conventions?

oh you're right, it should

supersaiyansubtlety (approved):

IotaBread (approved):

PR #538, opened 2024-01-22
Some Registry Names Refactoring

FirstMegaGame4:
# Not accurate with the current changes. Please read PR comments.

The current `Registries#REGISTRY` field seems to have a weird name (from my pov). Since it's containing all registries, I would say to use the same mapping name as Yarn, and refactor it to `Registries#REGISTRIES`.

It may be also a good thing to rename `Registries#MUTABLE_REGISTRY` to `Registries#ROOT` (also a Yarn mapping name) or maybe `Registries#ROOT_REGISTRY`? Since it's litterally using the `Registries#ROOT_REGISTRY_KEY`, so if these changes are agreed then I'll also change that name in the pull request.

Of course if you don't want these changes you can just close the pull request.


OroArmor (approved):

FirstMegaGame4:
okay so I'll also add the other mapping name change

Pyrofab:
I'd say the singular name makes sense ? It's just like every field in the `Registries` class - you have the "block registry", the "fluid registry", the "cat variant" registry, and the "registry registry"

FirstMegaGame4:
> I'd say the singular name makes sense ? It's just like every field in the `Registries` class - you have the "block registry", the "fluid registry", the "cat variant" registry, and the "registry registry"

It's a bit more complex.
The current registry you're talking about already exists, it's the `Registries#MUTABLE_REGISTRY` (which I renamed to `Registries#ROOT_REGISTRY` and explained why).

So since `Registries#REGISTRY` is just the same thing as `Registries#MUTABLE_REGISTRY`, I could see it more like a field that indicates clearly that it contains all the other ones, which means that turning it into the plural form is more logical and makes no issues since the singular name is still here.

Pyrofab:
Right, it seems to be a somewhat annoying case arising from mojang's registry implementation. I think adding "root" is a good change, makes it clear what we are talking about ; although I would probably consider the immutable one as the "canonical one", which should likely take the name `ROOT_REGISTRY` (or maybe `ROOT` for adequation with the key). The mutable one is private and I'd count it as an implementation detail - maybe `MUTABLE_ROOT_REGISTRY` would be enough ?

FirstMegaGame4:
It seems to be a good idea but the issue here is that the `ROOT` name is used by Yarn for the private mutable one. I don't think it's a good idea to confuse people.

supersaiyansubtlety:
I don't think avoiding confusion with yarn mappings should be considered as a reason not to name something.

FirstMegaGame4:
> Right, it seems to be a somewhat annoying case arising from mojang's registry implementation. I think adding "root" is a good change, makes it clear what we are talking about ; although I would probably consider the immutable one as the "canonical one", which should likely take the name `ROOT_REGISTRY` (or maybe `ROOT` for adequation with the key). The mutable one is private and I'd count it as an implementation detail - maybe `MUTABLE_ROOT_REGISTRY` would be enough ?

I'd say that `ROOT` and `INTERNAL_ROOT` can be two pretty good names. Giving a name based on its class type is a bit strange.

FirstMegaGame4:
Also maybe change `ROOT_REGISTRY_KEY` to `ROOT_KEY`? not sure about that

supersaiyansubtlety:
I think the existing `REGISTRY` is correct for the immutable view. 
The plurality matches that of the other immutable registry access fields in `Registries` and describes what it contains. 

If anything I think the mutable registry could be renamed to `ROOT` to match its key.

EDIT: to address https://github.com/QuiltMC/quilt-mappings/pull/538#issuecomment-1904535885

I think the public immutable one should be the one to follow the convention of the other registry fields in `Registries` because they're also public and immutable.

FirstMegaGame4:
> I think the existing `REGISTRY` is correct for the immutable view. The plurality matches that of the other immutable registry access fields in `Registries` and describes what it contains.
> 
> If anything I think the mutable registry could be renamed to `ROOT` to match its key.
> 
> EDIT: to address [#538 (comment)](https://github.com/QuiltMC/quilt-mappings/pull/538#issuecomment-1904535885)
> 
> I think the public immutable one should be the one to follow the convention of the other registry fields in `Registries` because they're also public and immutable.

I do not think that following the convention is a good think is that case.
As a QM user, I made this PR because the last time I used the `Registries` class, it took much too much time to understand why `REGISTRY` was actually the field holding the root registry. `ROOT`, or the Yarn name `REGISTRIES` are two much better options for understanding.

supersaiyansubtlety:
Hm, I don't see where the confusion comes from (I've not had to interact with these much with my mods) but I'll take your word for it.
I don't think `REGISTRIES` is an improvement, so I guess the current PR `ROOT` and `INTERNAL_ROOT` work.

supersaiyansubtlety (approved):

FirstMegaGame4:
> Hm, I don't see where the confusion comes from (I've not had to interact with these much with my mods) but I'll take your word for it. I don't think `REGISTRIES` is an improvement, so I guess the current PR `ROOT` and `INTERNAL_ROOT` work.

It probably comes from the fact that when you're looking for a registry of registries, you're either searching for the field that holds the `REGISTRIES` or for the field that holds the `ROOT` of all registries. So when you only see a `REGISTRY` field it sounds a bit strange.

Pyrofab (approved):

eristhea (approved):

IotaBread (approved):

FirstMegaGame4:
is there something missing here?

PR #541, opened 2024-01-31
Fix crafter screen mappings

IotaBread:
What's named `CrafterScreen` in the current mappings is actually the screen handler


supersaiyansubtlety (approved):

eristhea (approved):
brilliant!

PR #562, opened 2024-03-14
Update github actions workflow steps

pyrox0:
This removes previously deprecated NodeJS 16 actions, see https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/ for more info.

Specifically, I validated that each of these version bumps contains no functional changes, only Node 16 -> 20 version bumps. This should remove the deprecation warnings under every actions run.

I did not update `actions/upload-artifact` from v3 to v4, as that does contain functional changes. This should maybe be investigated, however, as it does contain significant performance increases(github claims in [their v4 release post](https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/) that it's a 10x increase), but it does change things and is not backwards-compatible with v3. If any tooling relies on v3 or its behavior, it would have to be adjusted.

I'd be glad to answer any questions about these changes if needed!

### DIFF
--- a/.github/actions/diff-uploader/action.yml
+++ b/.github/actions/diff-uploader/action.yml
@@ -11,5 +11,5 @@ outputs:
   result:
     description: 'Result of the action'
 runs:
-  using: 'node16'
+  using: 'node20'
   main: 'dist/index.js'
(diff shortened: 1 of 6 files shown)


IotaBread (approved):
I think the actions in `.github/actions` are also outdated so it'd be nice if you could also update those

eristhea (approved):

pyrox0:
Updated the above mentioned actions!

PR #568, opened 2024-03-25
CI Updates

pyrox0:
This PR does 2 things:
* A `pr-manager` action that checks FCP dates, notifies the mappings team if a PR is done in FCP, and also ensures all needed labels are assigned to a PR.
* Updates the javascript dependencies of the `update-base` and `diff-uploader` actions. This also solves some warnings that appear under running those actions.


.github/actions/pr-manager/index.js
```diff
@@ -0,0 +1,230 @@
...
+const smallLabel = "s: small";
+const medLabel = "s: medium";
+const largeLabel = "s: large";
+const newLabel = "T: new";
+const refactorLabel = "T: refactor";
+const docLabel = "T: documentation";
+const toolLabel = "T: toolchain";
+
+// If someone is added to the Mappings team, add their Github username here.
+const mappingsTeam = [
```
eristhea: since Quilt teams are granular, it should be possible to get the mappings team from the API instead of manually maintaining it

eristhea (commented):
no more !!remindme !

.github/actions/pr-manager/index.js
pyrox0: Updated this!

eristhea (approved):
I'm not a web dev but this looks fantastic!

.github/actions/pr-manager/index.js
```diff
@@ -0,0 +1,233 @@
...
+    };
+}
+
+// Ensure the FCP label is on any PRs that are in FCP.
+// If it's there, no need to do anything.
+async function ensureFCPLabel(labels, num) {
+    // If FCP label doesn't exist, add it by calling the API.
+    if (!labels.includes(fcpLabel)) {
+        await github.issues.addLabels({ ...baseParameters, issue_number: num, labels: [fcpLabel] });
+    } else { };
```
IotaBread: why this else?

.github/actions/pr-manager/index.js
```diff
@@ -0,0 +1,233 @@
...
+            if (comment.body.includes("start-fcp")) {
+                commentStatus = await checkFCPComment(comment);
+                if (commentStatus == 1) {
+                    return;
+                }
+                // Ensure we have the FCP label.
+                await ensureFCPLabel(labels, pull.number);
+                // Check the FCP time.
+                fcpTime = await checkFCPTime(comment, pull);
+                if (fcpTime <= 1) {
```
IotaBread: Empty if

.github/actions/pr-manager/index.js
pyrox0: I missed that I had added that when doing my final code check, removed.

.github/actions/pr-manager/index.js
pyrox0: Removed this as well.

pyrox0:
Taking this out of FCP since there's some improvements I need to make to it.

OroArmor (changes requested):
Looks good, but a lot of the information is within the log and not a comment. Having a comment or something would be much nicer. 

Also I could have missed it, but is there a way to cancel an FCP

moehreag (approved):
From what I currently understand of gh actions and js this looks good (Oro's request is a good idea though).

pyrox0:
Closing per what I said in the quilt mappings discord thread, someone else can pick this up if they want.

PR #575, opened 2024-04-02
spellcheck dictionary update - remove "armour" and "initialise"

eristhea:
commit: https://github.com/ix0rai/qm-base-allowed-wordlist/commit/2093ffe06086bbb1dfe35428975dbbc3248aa89f


supersaiyansubtlety (approved):

pyrox0 (approved):

sylv256:
I thought we used Standard American English conventions

supersaiyansubtlety:
> I thought we used Standard American English conventions

yes, armour and initialise are both European spellings, so they were removed

eristhea:
(when something is removed from the dictionary, it means that a mapping containing that word will throw an error and fail mapping lint)

PR #576, opened 2024-04-04
Rescue rai from mappings nightmare

EnnuiL:
- SystemToast.Type (unmapped) has been refactored to SystemToast.Id since Type is a more misleading name while Id more accurately states the intention as a sort of toast id that differentiates toasts from each other
- MapIcon still needs to be completely refactored into MapDecoration (or the vice-versa..?)
- accumulatedScrollX/Y has been decopypasted from Mojmap; it's now scrollDeltaX/Y, especially since there is no accumulation going on here (unlike cursorDeltaX/Y, although after some discussion, that name is alright)


src/constants/java/org/quiltmc/quiltmappings/constants/LlamaVariants.java
OroArmor, line 1: Why were these deleted?

EnnuiL: > Why were these deleted?

Made obsolete by Minecraft itself (Minecraft has LlamaVariants, setVariant asks for LlamaVariants as well; same case for parrots)

OroArmor: Thats right. Anything left in the constants project?

EnnuiL: > Thats right. Anything left in the constants project?

Only the package-info.java, really

OroArmor: Yikes. that needs to be fixed

eristhea:
my knight in shining armour

mappings/net/minecraft/client/gui/screen/BackupPromptScreen.mapping
```diff
@@ -1,12 +1,17 @@
 CLASS net/minecraft/unmapped/C_elvupmqi net/minecraft/client/gui/screen/BackupPromptScreen
 	FIELD f_bbxlfnbn showEraseCacheCheckbox Z
-	FIELD f_khazxlgf subtitle Lnet/minecraft/unmapped/C_rdaqiwdt;
+	FIELD f_fewgdmzl runnable Ljava/lang/Runnable;
```
eristhea: this runnable is triggered when the ESC key is pressed and has code that closes resources
I'd call it something like `onCanceled`
```suggestion
	FIELD f_fewgdmzl onCanceled Ljava/lang/Runnable;
```

mappings/net/minecraft/client/gui/screen/BackupPromptScreen.mapping
```diff
@@ -1,12 +1,17 @@
...
+	FIELD f_khazxlgf description Lnet/minecraft/unmapped/C_rdaqiwdt;
 	FIELD f_mblzcmck id I
 	FIELD f_okzzeovr wrappedText Lnet/minecraft/unmapped/C_ogfbkizf;
 	FIELD f_vwuzyxro eraseCacheCheckbox Lnet/minecraft/unmapped/C_lvxxsthn;
 	FIELD f_wwwbtzmt callback Lnet/minecraft/unmapped/C_elvupmqi$C_agdgbivq;
 	METHOD <init> (Ljava/lang/Runnable;Lnet/minecraft/unmapped/C_elvupmqi$C_agdgbivq;Lnet/minecraft/unmapped/C_rdaqiwdt;Lnet/minecraft/unmapped/C_rdaqiwdt;Z)V
+		ARG 1 runnable
```
eristhea:
```suggestion
```
can be removed and automapped

mappings/net/minecraft/client/toast/SystemToast.mapping
```diff
@@ -16,24 +31,52 @@ CLASS net/minecraft/unmapped/C_vksdgrxp net/minecraft/client/toast/SystemToast
...
 		ARG 0 client
+		ARG 1 id
 		ARG 2 title
 		ARG 3 description
+	METHOD m_yghdjgwj hide ()V
+	METHOD m_zfyvjxrz (Lnet/minecraft/unmapped/C_sedilmty;Lnet/minecraft/unmapped/C_eqfjupcc;J)Lnet/minecraft/unmapped/C_zqhkdxaf$C_sakiolwc;
+		ARG 1 graphics
 	METHOD m_zstagntz getTextAsList (Lnet/minecraft/unmapped/C_rdaqiwdt;)Lcom/google/common/collect/ImmutableList;
 		ARG 0 text
+	CLASS C_ozahoshp Id
```
eristhea, line 71: I still prefer the `Type` name here. your argument for the new name doesn't really make sense to me
> a sort of toast id that differentiates toasts from each other
`type` is also a word that implies differentiation? I may be dense but I don't really get the refactor

eristhea (changes requested):
looks great!

mappings/net/minecraft/client/toast/SystemToast.mapping
eristhea: to expand my position a little, usually `id` is something more internal while `type` is something more user-facing

mappings/net/minecraft/client/gui/screen/BackupPromptScreen.mapping
EnnuiL: ah fair, although unfortunately the convention favors me and my beloved cancelled, not your canceled :p

mappings/net/minecraft/client/gui/screen/BackupPromptScreen.mapping
EnnuiL: Hm, am not sure about how to feel about relying on automappings; I'll keep it but with the correct mapping

mappings/net/minecraft/client/gui/screen/BackupPromptScreen.mapping
eristhea: oh lol I corrected myself from the proper spelling because I thought american was 1 l

mappings/net/minecraft/client/toast/SystemToast.mapping
EnnuiL: The issue of Type is that it implies that it can be recycled between toasts of a certain type while in fact? It's more of an identifier for one specific toast; The SystemToast.Type name had made me mistakenly use a vanilla toast "type" for a modded toast when in fact it shouldn't have been used that way; This is why I kinda resist against `Type`;

I guess this refactor could be reverted though, it's just that yeah, the original name bothers me; it feels misleading

eristhea: oh girl i did the exact same thing as you in my mod is that not what i'm supposed to do

EnnuiL: > oh girl i did the exact same thing as you in my mod is that not what i'm supposed to do

You are not! You are supposed to create your own `SystemToast.<Thing>` and use it for toasts; If you use a Vanilla one? If the Vanilla one and your toast attempts to be added at the same time? One's gonna overwrite the other since they share the same id! Hence why am proposing the Mojmap-inspired `SystemToast.Id` (moj's way more verbose though): because it is effectively the id of a toast, and it is the method which allows toasts to be differentiated when added

supersaiyansubtlety: Could you add a commend to `Toast#getId` clarifying that toasts with the same id will override one another?

EnnuiL: Sure! I can Javadoc something; not sure if at `Toast#getId`, but I can try something

mappings/net/minecraft/client/toast/ToastManager.mapping
```diff
@@ -6,7 +6,7 @@ CLASS net/minecraft/unmapped/C_eqfjupcc net/minecraft/client/toast/ToastManager
 		ARG 1 client
 	METHOD m_glbfkboq getToast (Ljava/lang/Class;Ljava/lang/Object;)Lnet/minecraft/unmapped/C_zqhkdxaf;
 		ARG 1 toastClass
-		ARG 2 type
+		ARG 2 id
 	METHOD m_lnyjvhnt clear ()V
 	METHOD m_lsmszfie getNotificationDisplayTime ()D
```
supersaiyansubtlety, line 14:
```suggestion
	METHOD m_lsmszfie getNotificationDisplaySeconds ()D
```
I think either this^ or `SystemToast.Id.displayDuration` -> `SystemToast.Id.displayMilis`

supersaiyansubtlety (changes requested):

mappings/net/minecraft/client/toast/ToastManager.mapping
EnnuiL: I actually like your latter suggestion! Lemme see if I can apply it

mappings/net/minecraft/client/toast/SystemToast.mapping
EnnuiL: I opted for a Javadoc on SystemToast.Id itself

supersaiyansubtlety (approved):

eristhea (approved):

mappings/net/minecraft/client/toast/SystemToast.mapping
supersaiyansubtlety: Hm, doesn't it apply to any object returned as an id from `Toast#getId`, including the default `ID` in `Toast`?

I think it would make more sense to put it there, with an `@see` pointing to `SystemToast.Id`.

mappings/net/minecraft/client/toast/Toast.mapping
```diff
@@ -1,9 +1,10 @@
 CLASS net/minecraft/unmapped/C_zqhkdxaf net/minecraft/client/toast/Toast
-	FIELD f_msurgqxf TYPE Ljava/lang/Object;
+	FIELD f_msurgqxf ID Ljava/lang/Object;
```
supersaiyansubtlety:
```suggestion
	FIELD f_msurgqxf DEFAULT_ID Ljava/lang/Object;
```

supersaiyansubtlety (changes requested):

mappings/net/minecraft/client/toast/Toast.mapping
EnnuiL: Mojmap actually suggests that this weird Id is actually the id that is used on toasts that don't have the same system that `SystemToast` has and only really has one id that is used! Such as `RecipeToast`; It calls it `NO_TOKEN`, but i don't think it's a good name?

EnnuiL: oh well, placeholder name for now..? there should be a better name though

supersaiyansubtlety: Do non-`SystemToast` toasts not override each other (afaict they all use this default id)?

I thought they did based on how `ToastManager#getToast` works.

EnnuiL: So, apparently they avoid this by using different classes alongside the ids? I still have to look at this more closely (something that is now impossible thanks to 24w14a branch being broken :p)

supersaiyansubtlety: But then they still override toasts of the same class (because they all have the same id).

Maybe this should be `NO_ID` with javadoc
> The default identifier used by toast classes for which only one instance is displayed at a time.

And a javadoc at `getId` could be something like
> Toasts with the same class with different identifiers can co-exist, while toasts sharing the same class and identifier one will overwrite each other.

(mostly copied from your javadoc)

EnnuiL: Ah fair; this matches with Mojang's intention of `NO_TOKEN` perfectly while not being confusing at the same time; I'll investigate this tomorrow, since a broken Enigma complicates things, and yeah, if I have to workaround, I'd rather work around it tomorrow morning :p

(btw, `getId` is once again a silly target, I'd find an another more appropriate method; clarifying Javadocs, however, are the correct path for this)

EnnuiL: I applied the `NO_ID` suggestion; however, the Javadoc should be done on another tackle on unmapped Toast stuff; it needs to be thought about carefully, but it should be worth it

supersaiyansubtlety (approved):

PR #578, opened 2024-04-10
fix buildscript for 24w14a

eristhea:
- update java to 21
- update dependencies
- unpick is still broken


build.gradle
```diff
@@ -12,6 +12,10 @@ buildscript {
 			name "Fabric Repository"
 			url 'https://maven.fabricmc.net'
 		}
+		// remove when ASM fix is released
+		maven {
+			url = "https://repository.ow2.org/nexus/content/repositories/snapshots/"
```
IotaBread, line 17: Is it needed at buildscript level?

IotaBread (approved):

build.gradle
eristhea: yes for running :mappings

PR #559, opened 2024-03-11
item components

eristhea:
based on #557, as I needed that work to make sense of a lot of this
names lean towards Mojang's names a lot more than Yarn's, as I find them more intuitive

PR #581, opened 2024-04-15
complete components, text, and village

eristhea:
again, straightforward.
refactor: `TooltipContext` -> `TooltipType`
I found a class in `Item` that much better matches the context name, and so this needed a new name. I was considering both `TooltipProperties` and `TooltipType`, and chose `Type`, though it could go either way imo.


mappings/net/minecraft/client/item/TooltipType.mapping
supersaiyansubtlety, line 1: How about `TooltipConfig`?

mappings/net/minecraft/component/type/DyedColorComponent.mapping
supersaiyansubtlety, line 1: I think all `showInTooltip` should be `shownInTooltip` in this class, `UnbreakableComponent`, and `AttributeModifiersComponent` (this is how it is in `ItemEnchantmentsComponent`).

see https://github.com/QuiltMC/quilt-mappings/pull/559#discussion_r1525123993

supersaiyansubtlety (commented):
A couple suggestions

mappings/net/minecraft/client/item/TooltipType.mapping
eristhea: that works with me!

eristhea:
both suggestions accepted!

supersaiyansubtlety (approved):

OroArmor (approved):

PR #591, opened 2024-05-11
Fixed range method mappings.

supersaiyansubtlety:
- **Fixed** the maximum paramaters of the two `range...` methods each indicating the opposite they should have been of inclusive/exclusive.
- Added javadocs clarifying what the bounded methods do.

Note that `range` throws an exception with the message
> bound - origin is non positive

I'll update param names to match this if that's preferred, but I haven't yet because I find `min`/`max` much more more intuitive.


mappings/net/minecraft/util/random/RandomGenerator.mapping
```diff
@@ -4,7 +4,10 @@ CLASS net/minecraft/unmapped/C_rlomrsco net/minecraft/util/random/RandomGenerato
 	COMMENT <p> This interface should be used instead of {@link java.util.Random} for random number generation.
 	FIELD f_dhenmqup GAUSSIAN_SPREAD_FACTOR D
 	METHOD m_afzjxjzq nextInt (I)I
-		ARG 1 bound
+		COMMENT @return a random value between {@code 0} (inclusive) and {@code bound} (exclusive)
+		COMMENT @throws IllegalArgumentException if {@code bound} isn't positive
```
eristhea: I like the renames, make sure to swap this javadoc to the `max` name!

eristhea (approved):

mappings/net/minecraft/util/random/RandomGenerator.mapping
supersaiyansubtlety: oops, fixed

PR #593, opened 2024-06-01
completed recipe and inventory

supersaiyansubtlety:
I went pretty literal for the `BrewingRecipeRegistry` top/bottom stuff, let me know if you prefer mojmap or yarn names, or if you have other suggestions.

I also chose "symmetric" over "symmetrical" because it's the same thing but shorter, and omitted "horizontal" from "symmetric"/"mirror" except for `isHorizontallySymmetric` because I think that's the most important place to be explicit about it.

### DIFF
--- a/mappings/net/minecraft/recipe/RecipeGridAligner.mapping
+++ b/mappings/net/minecraft/recipe/RecipeGridAligner.mapping
@@ -8,7 +8,7 @@ CLASS net/minecraft/unmapped/C_wuyfqgmj net/minecraft/recipe/RecipeGridAligner
 		ARG 6 amount
 	METHOD m_vnshednj collectToInput (Ljava/lang/Object;IIII)V
 		ARG 1 itemId
-		ARG 2 iSlot
+		ARG 2 slotIndex
 		ARG 3 amount
 		ARG 4 x
 		ARG 5 y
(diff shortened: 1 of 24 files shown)


mappings/net/minecraft/recipe/RecipeGridAligner.mapping
```diff
@@ -6,3 +6,9 @@ CLASS net/minecraft/unmapped/C_wuyfqgmj net/minecraft/recipe/RecipeGridAligner
 		ARG 4 recipe
 		ARG 5 inputs
 		ARG 6 amount
+	METHOD m_vnshednj collectToInput (Ljava/lang/Object;IIII)V
+		ARG 1 itemId
+		ARG 2 iSlot
```
eristhea: why not `slotIndex`?

eristhea (approved):
looks fantastic, just one comment!

mappings/net/minecraft/recipe/RecipeGridAligner.mapping
supersaiyansubtlety: Oh, that's something I do in my own code because it's brief and works when there are multiple indices (`jSlot`), forgot it's not standard for QMap.

I'll change it.

PR #594, opened 2024-06-02
[1.20.6] CustomPayload.Id -> Type

TheGlitch76:
This class is not just an id, it associates an id with a specific CustomPayload class (a la EntityType).

Provided for 1.20.6 because that's what QSL is on and i'd like to use this in some api class names, but I will live without it until 1.21 if it's not possible to refactor an old version


eristhea:
(note: glitch mentioned in discord that this PR was based on a misunderstanding of the code)

PR #602, opened 2024-06-15
Mention the versions API in readme.

mk-pmb:
Thanks again, @FirstMegaGame4 , for your solution over in https://github.com/QuiltMC/quiltmc.org/issues/229 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,14 @@ maven { url = "https://maven.quiltmc.org/repository/release/" }
 QM versions are named after the Minecraft version they target, followed by `+build.x`,
 with `x` being a number that increments with each new build of QM for that Minecraft version.
 For example, `1.18.2+build.1` is the first build of QM for Minecraft `1.18.2`. You can see additional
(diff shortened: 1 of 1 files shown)


supersaiyansubtlety (approved):

eristhea (approved):
would you mind also updating the link to the import tool to be the one on quiltmc.org?

mk-pmb:
Amended it to point to the query for a specific Minecraft version, in hopes that this will reduce server load.

mk-pmb:
I tried to update the import tool link, but I cannot determine whether the link caption "LambdAurora's import tool" is still correct, as the new target seems to not mention "LambdAurora".

supersaiyansubtlety:
iirc the new one is by simpledemon
but since it's on the quilt site it's probably ok to just say "quilt import tool"

mk-pmb:
I updated the commit and also separated the URLs from the text in order to reduce future diff noise if the URLs change again.

⚠ An attribution has been removed. Maintainers should verify this before merging, especially if the tool's license may require attribution.

eristhea (approved):
I didn't know you could do that in markdown

OroArmor (approved):

IotaBread (approved):

PR #603, opened 2024-06-26
Networking mappings

TheGlitch76:
This contains a couple refactors:
`NetworkState` -> `NetworkPhase` -- we already refer to the different protocol phases as "phases" elsewhere in the code, this was an outlier.
`ConnectionPhase` (and affiliated classes) -> `NetworkPhaseProtocol` -- Had to be renamed to not clash confusingly with `NetworkPhase`. `NetworkPhaseProtocol` represents the serde for all the individual packets that make up the protocol of the phase. This distinction isn't made by Mojang (they actually just call everything the "protocol" and use `ConnectionProtocol` in lieu of what I'm calling `NetworkPhase`) but it seems to fit with how the code is structured.


mappings/net/minecraft/network/listener/PacketListener.mapping
```diff
@@ -4,8 +4,13 @@ CLASS net/minecraft/unmapped/C_tkgdawqw net/minecraft/network/listener/PacketLis
 	COMMENT
 	COMMENT <p>Its listener methods will be called on the netty event loop than the
```
supersaiyansubtlety:
```suggestion
	COMMENT <p>Its listener methods will be called on the netty event loop, then the
```
is that right?
or is `than` supposed to be `from`?

supersaiyansubtlety (approved):

eristhea (approved):
looks great!

IotaBread (approved):

PR #617, opened 2024-07-23
Add `findDuplicateMappingFiles` task

supersaiyansubtlety:
Adds `findDuplicateMappingFiles` gradle task that checks for mapping files that map the same class.

Useful when merging/rebasing.
It just matches the first line of each file against ~~`"^CLASS net/minecraft/unmapped/C_\\w+"`~~ and if it finds any files with the same one it throws an exception listing them.

edit: it matches against this now:
```regex
"^CLASS (?:net/minecraft|com/mojang/blaze3d)/(?:\\w+/)*\\w+(?= )"
```

Also makes `mappingLint` depend on `findDuplicateMappingFiles`.

Edit: also removes current duplicate mapping files so the build run can complete.

### DIFF
--- a/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
+++ b/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
@@ -21,8 +21,8 @@ public abstract class FindDuplicateMappingFilesTask extends DefaultTask {
     public static final String TASK_NAME = "findDuplicateMappingFiles";
 
     private static final Logger LOGGER = Logging.getLogger(FindDuplicateMappingFilesTask.class);
-    private static final Pattern MINECRAFT_CLASS = Pattern.compile("^CLASS net/minecraft/(?:\\w+/)*\\w+ ");
-    private static final Pattern BLAZE_CLASS = Pattern.compile("^CLASS com/mojang/blaze3d/(?:\\w+/)*\\w+ ");
+    private static final Pattern MINECRAFT_CLASS = Pattern.compile("^CLASS net/minecraft/(?:\\w+/)*\\w+(?= )");
+    private static final Pattern BLAZE_CLASS = Pattern.compile("^CLASS com/mojang/blaze3d/(?:\\w+/)*\\w+(?= )");
 
     @InputDirectory
     public abstract DirectoryProperty getMappingDirectory();
--- a/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
+++ b/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
@@ -52,12 +52,12 @@ public abstract class FindDuplicateMappingFilesTask extends DefaultTask {
                                 },
                                 () -> mapNoClassFiles.add(mappingFile)
                             );
-
-                            if (!mappingFile.toString().endsWith(".mapping")) {
-                                wrongExtensionFiles.add(mappingFile);
-                            }
                         } else {
                             emptyFiles.add(mappingFile);
+                        }
+
+                        if (!mappingFile.toString().endsWith(".mapping")) {
+                            wrongExtensionFiles.add(mappingFile);
                         }
                     } catch (IOException e) {
                         throw new GradleException("Unexpected error accessing mapping file", e);
@@ -119,8 +119,8 @@ public abstract class FindDuplicateMappingFilesTask extends DefaultTask {
             errorMessages.add(message);
 
             LOGGER.error("Found {}!", message);
-            for (final File mapNoClassFile : mapNoClassFiles) {
-                LOGGER.error("\t{}", mapNoClassFile);
+            for (final File wrongExtensionFile : wrongExtensionFiles) {
+                LOGGER.error("\t{}", wrongExtensionFile);
             }
         }
 
--- a/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
+++ b/buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
@@ -21,8 +21,8 @@ public abstract class FindDuplicateMappingFilesTask extends DefaultTask {
     public static final String TASK_NAME = "findDuplicateMappingFiles";
 
     private static final Logger LOGGER = Logging.getLogger(FindDuplicateMappingFilesTask.class);
-    private static final Pattern MINECRAFT_CLASS = Pattern.compile("^CLASS net/minecraft/(?:\\w+/)*\\w+(?= )");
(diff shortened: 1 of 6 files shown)


supersaiyansubtlety:
~~Output looks like this~~ (these are actual duplicates currently on the 1.21 branch):
edit: see new logging below
<pre>
> Found classes mapped by multiple files!
      CLASS net/minecraft/unmapped/C_zjsuejaf is mapped by:
          C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\datafixer\fix\AttributeIdFix.mapping
          C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\datafixer\fix\AttributeModifierIdFix.mapping
      CLASS net/minecraft/unmapped/C_dxtlxcwy is mapped by:
          C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\datafixer\fix\RemoveEmptyItemInSuspiciousBlockFix.mapping
          C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\datafixer\fix\RemoveEmptyItemInBrushableBlockFix.mapping
</pre>

supersaiyansubtlety:
It would also be trivial to add checking for empty files, files that don't match the regex, and files that don't have the `mapping` extension. 
Do we care about those?

supersaiyansubtlety:
Went ahead and added those checks. 
It also checks non-obfuscated classes now and the logging is more similar to `MappingLintTask`.
The new logging:
<pre>
Found 1 class mapped by multiple files!
	CLASS net/minecraft/unmapped/C_xyxorzsk is mapped by:
		C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersion.mapping
		C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersion2.mapping2
Found 1 empty file!
	C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersionEMPTY.mapping3
Found 1 file not mapping a class!
	C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersionNO_CLASS.mapping
Found 2 files without the mapping extension!
	C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersion2.mapping2
	C:\Users\super\Projects\Minecraft\Modding\Quilt\quilt-mappings\mappings\net\minecraft\WorldVersionEMPTY.mapping3

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':findDuplicateMappingFiles'.
> Found 1 class mapped by multiple files, 1 empty file, 1 file not mapping a class, and 2 files without the mapping extension! See the log for details.
</pre>

supersaiyansubtlety:
```java
private static final Pattern MINECRAFT_CLASS = Pattern.compile("^CLASS net/minecraft/(?:\\w+/)*\\w+(?= )");
private static final Pattern BLAZE_CLASS = Pattern.compile("^CLASS com/mojang/blaze3d/(?:\\w+/)*\\w+(?= )");
```
these could be generalized to
```java
private static final Pattern ANY_CLASS = Pattern.compile("^CLASS (?:\\w+/)+\\w+(?= )");
```
idk if we want to allow classes outside of `net.minecraft` and `com.mojang.blaze3d` though

OroArmor:
Until a good reason comes, I think disallowing anything outside of those two root packages is a good idea.

buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
```diff
@@ -0,0 +1,161 @@
...
+import org.gradle.api.tasks.TaskAction;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+public abstract class FindDuplicateMappingFilesTask extends DefaultTask {
```
OroArmor, line 20: Could this be a `Checker` and not a full task?

OroArmor (changes requested):

buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
supersaiyansubtlety: Sure, just... let me figure out what a `Checker` is

OroArmor: Its part of the buildSrc, should be a bit cleaner.

supersaiyansubtlety: It looks like `Checker`s only see one mapping at a time, so I'd have to ~~maintain `allMappings` and `duplicateMappings` in static fields, or~~ add methods to `Checker` to allow attaching things to a task run.
It also doesn't look like they expose the actual mapping file, so I couldn't do the extension check without adding methods to `Checker` for it.

+On the other hand it would mean I could put the "empty" and "extension" checks in separate `Checker`s.

Is that preferable to a separate task?

supersaiyansubtlety: I'll start implementing it and see how it looks

OroArmor (approved):

buildSrc/src/main/java/quilt/internal/tasks/lint/FindDuplicateMappingFilesTask.java
supersaiyansubtlety: conclusion: it can't be a `Checker` because enigma doesn't create multiple entries if multiple files map a class

eristhea (approved):
fantastic work!

OroArmor (changes requested):
can we have the `check` task depend on this like the `MappingsLint` task?

supersaiyansubtlety:
> can we have the `check` task depend on this like the `MappingsLint` task?

Sure; instead of `MappingsLint`'s dependency on `findDuplicateMappingFiles`, or in addition to it?

OroArmor:
Oh! I missed that! Looks good!

OroArmor (approved):

OroArmor:
@ix0rai with this, do we want to enforce that all PRs are up to date before merging (like QSL)? it'll make merging multiple PRs in a row annoying, but will make sure that this is caught in a PR and not once 3 branches are merged at once

eristhea:
> @ix0rai with this, do we want to enforce that all PRs are up to date before merging (like QSL)? it'll make merging multiple PRs in a row annoying, but will make sure that this is caught in a PR and not once 3 branches are merged at once

I think we should, it's probably possible to automate checking them using actions and merge queue

PR #640, opened 2024-09-17
fix+clarify ChunkTickScheduler, complete n.m.world.tick

supersaiyansubtlety:
When adding a block tick scheduler to one of my mods I was dismayed to find `ChunkTickScheduler` had three fields that seemed very similar:
```java
Queue<OrderedTick<T>> tickQueue;
List<Tick<T>> ticks;
Set<OrderedTick<?>> queuedTicks;
```
I was also surprised that `void disable(long time)` was the method responsible for making sure deserialized ticks actually get scheduled.

I've renamed them:
```java
Queue<OrderedTick<T>> scheduledTicks;
List<Tick<T>> initialTicks;
Set<OrderedTick<?>> allTicks;
//...
void scheduleInitialTicks(long time)
```

I also added some javadocs clarifying how `initialTicks` works and renamed things to more consistently use "schedule".

---

Now also completes `n.m.world.tick`


OroArmor:
You should take a look at the new world interface that exists now

supersaiyansubtlety:
> You should take a look at the new world interface that exists now

`C_adoatlrf` ?

that does look tick-y

OroArmor:
probably. `ScheduledTickAccess` in MM

OroArmor (approved):

eristhea (approved):
looks good, great refactors!

PR #645, opened 2024-09-28
gradle cleanup

### LINKED ISSUE #492: Use gradle home cache for Minecraft libraries
Storing the Minecraft libraries in the home cache would reduce wait times and disk usage greatly when managing two or more clones of the repository, instead of downloading them on each copy of QM.

supersaiyansubtlety:
### Tasks
- [x] the thisening
- [x] the finalening
- [x] the line wrappening
- [x] the abstractening
- [x] the registerening
- [x] move all custom tasks to plugins
- [x] move (most) task configuration to plugins (makes it easier to see dependencies and makes tasks more reusable)
- [x] eliminate `buildSrc` referencing `build.gradle` things
- [x] eliminate project access in task actions
- [x] refine inputs/outputs and dependencies
- [x] move more things to `QuiltMappingsExtension`
- [x] separate `QuiltMappingsPlugin`'s contents into several plugins
- [x] wait for #659 which fixes tests, make sure tests still pass
- [x] resolve a few remaining questions
- ~~replace `TasksTest` and depend on gradle test kit~~ maybe in a followup PR
- ~~apply a checkstyle to `buildSrc`?~~ maybe in a followup PR

## Summary

renames `MappingsPlugin`/`mappings-logic` -> `QuiltMappingsPlugin`/`quilt-mappings`
renames `MappingssExtension`/`mappings` -> `QuiltMappingsExtension`/`quiltMappings`
splits `QuiltMappingsPlugin` into:
- QuiltMappingsBasePlugin
- MinecraftJarsPlugin
- MapMinecraftJarsPlugin
- MapV2Plugin
- MapIntermediaryPlugin
- ProcessMappingsPlugin
- EnigmaMappingsPlugin
- TargetDiffPlugin

all implement `MappingsProjectPlugin` and are applied by `QuiltMappingsPlugin`

eliminates `FileConstants`:
- some common directories are in `MappingsProjectPlugin`
- files that were the output of one task are now defined where that task's output is set and other tasks take that output as input

adds several extensions that expose objects created by their plugins
extensions are accessed via plugin instances which ensures extension have been created

no use of `tasks.named(...)` to access custom tasks:
- instead they're accessed via the provider returned by `tasks.register(...)`
- tasks used outside of the plugin they're defined in are accessed via `Tasks` objects obtained via extensions, obtained via plugin instances
- this makes sure required plugins are applied and the tasks are registered

ensures all custom tasks have a group
eliminates `MappingsTask` and `DefaultMappingsTask`
(some `MappingsTask` methods are replaced with helper methods in more specific task interfaces and plugins)

moves `Constants.MINECRAFT_VERSION` to `libs.versions.toml` and passes it to `QuiltMappingsExtension`
moves `Constants.MAPPINGS_VERSION` to direct `project.version` assignment in `build.gradle` and passes it to `QuiltMappingsExtension`
moves dictionary file url to `libs.versions.toml` and downloads it with an ivy repository hack
eliminates `downloadDictionaryFile` and passes the dictionary file to `mappingLint` in  `build.gradle`
saves `unpickVersion` in a `QuiltMappingsEntension` field and passes it to the constructors of tasks that require it
    (it's required at configuration time for some tasks so it can't be a property)

extracts a bunch of magic strings to interfaces in `constants` package
moves all task names to their classes and renames all `TASK_NAME` fields `<ACTUAL_NAME>_TASK_NAME`
adds javadocs to all task names that link the plugin that registers them
adds javadocs to each plugin with any `configurEach`es listing what they do (these can be annoying to track down)
adds javadocs to each class that's `configureEach`ed and to all their subclasses, linking the plugin that does the configuring (these can be really annoying to track down)

moves almost all outputs to `build/*`
moves `combineUnpickDefinitions`'s output to `build/mappings/`
eliminates custom `clean` task logic (everything is in `build/` and gets cleaned)

names `AbstractArchiveTask` outputs using its name composition methods
adds `ArtifactFileProducingTask` which mimics `AbstractArchiveTask` name composition (implemented by `CompressTinyTask`)
eliminates explicit classifiers for published artifacts in `build.gralde`

mapped providers are the solution to every problem

moves `unpick.json` to `unpick/` and unpick definitions to `unpick/definitions/`
moves `enimga_profile.json` and `simple_type_field_names.json5` to `enigma/`

eliminates `decompileClasspath` configuration -> passes `downloadMinecraftLibraries.librariesDir.fileTree`
replaces `CheckStuffExists/Matches` tasks with mapped providers

populates `unpickCli` configuration using included build `:unpick-holders`
splits `enigmaRuntime` configuration into `enigmaSwing` and `enigmaServer`, populates them using `:unpick-holders`

replaces `de.undercouch.gradle.tasks.download.DownloadAction` usage with `DownloadUtil` methods (which use `org.apache.commons.io.FileUtils::copyURLToFile`)
eliminates `de.undercouch:gradle-download-task` dependency

adds `generateDiff` task (Windows users may have to put `diff` on their PATH)

### Future
- use `generateDiff` in `gerenate-diff.yml` (pass output path to avoid magic string)
- move `namedSource/` and `namedTargetSource/`once `generateDiff` is used in `gerenate-diff.yml`
- update/replace licenser plugin: the last release is from 2021 and it's incompatible with gradle 9
- `@CachableTask` everything
- [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html)


supersaiyansubtlety:
~~Early reviewers beware: I will probably be rewriting history so any review progress on github may be lost.~~
Ready ~~-ish~~

buildSrc/src/main/java/quilt/internal/MappingsExtension.java
```diff
@@ -1,15 +1,54 @@
...
+    public static MappingsExtension get(Project project) {
+        return project.getExtensions().getByType(MappingsExtension.class);
+    }
+
+    protected abstract RegularFileProperty getEnigmaProfileFile();
+
+    public MappingsExtension(Project project) {
+        this.fileConstants = new FileConstants(project);
+
+        this.enigmaProfile = this.getEnigmaProfileFile()
```
lukebemish: `map`ed providers _are not cached_, and I assume this is used in several placed  -- so instead of having, say, a `public final Provider<EnigmaProfile> enigmaProfile`, have a `Property<EnigmaProfile>` and `set` it to the mapped provider, which _is_ cached

buildSrc/src/main/java/quilt/internal/MappingsPlugin.java
```diff
@@ -24,66 +31,277 @@
...
-        tasks.create(OpenGlConstantUnpickGenerator.TASK_NAME, OpenGlConstantUnpickGenerator.class);
-
-        tasks.create(GeneratePackageInfoMappingsTask.TASK_NAME, GeneratePackageInfoMappingsTask.class);
-        tasks.create(DownloadDictionaryFileTask.TASK_NAME, DownloadDictionaryFileTask.class);
-        final var mappingLintTask = tasks.create(MappingLintTask.TASK_NAME, MappingLintTask.class);
-        tasks.create(FindDuplicateMappingFilesTask.TASK_NAME, FindDuplicateMappingFilesTask.class, task -> {
-            task.getMappingDirectory().set(mappingLintTask.getMappingDirectory());
-            mappingLintTask.dependsOn(task);
+    public void apply(@NotNull Project project) {
+        final var ext = project.getExtensions().create("mappings", MappingsExtension.class, project);
```
lukebemish: Not _really_ a big deal, but the `Project` can just be `@Inject`ed here.

buildSrc/src/main/java/quilt/internal/tasks/EnigmaProfileConsumingTask.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static org.quiltmc.enigma_plugin.Arguments.SIMPLE_TYPE_FIELD_NAMES_PATH;
+
+public abstract class EnigmaProfileConsumingTask extends DefaultMappingsTask {
+    @Internal(
+        "An EnigmaProfile cannot be fingerprinted. " +
+            "Up-to-date-ness is ensured by getSimpleTypeFieldNamesFiles and its source, " +
+            "MappingsExtension::getEnigmaProfileFile."
+    )
+    public abstract Property<EnigmaProfile> getEnigmaProfile();
```
lukebemish: This is enough for build cache, but won't get you to config cache -- _all_ task properties need to be serialize-able or transient (in which case they may not exist at task runtime), `@Internal` or `@Input`s or `@Output`s or whatever. Question for your consideration: is there ever more than a single `EnigmaProfile` created by this plugin, at all? If the answer is no, use a build service!

buildSrc/src/main/java/quilt/internal/tasks/EnigmaProfileConsumingTask.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Collection;
+import java.util.stream.Stream;
+
+import static org.quiltmc.enigma_plugin.Arguments.SIMPLE_TYPE_FIELD_NAMES_PATH;
+
+public abstract class EnigmaProfileConsumingTask extends DefaultMappingsTask {
+    @Internal(
+        "An EnigmaProfile cannot be fingerprinted. " +
+            "Up-to-date-ness is ensured by getSimpleTypeFieldNamesFiles and its source, " +
+            "MappingsExtension::getEnigmaProfileFile."
```
lukebemish: Is `MappingsExtension::getEnigmaProfileFile` ever actually marked/added as a task input? Or is the only stuff from it that can affect stuff the simpleTypeFieldNamesFiles?

buildSrc/src/main/java/quilt/internal/tasks/decompile/DecompileTask.java
```diff
@@ -23,89 +24,69 @@
...
+        if (this.methodJavadocProvider != null) {
+            decompiler.withMethodJavadocProvider(this.methodJavadocProvider);
         }
 
-        decompiler.decompile(getInput().getAsFile().get(), getOutput().getAsFile().get(), options, libraries);
+        decompiler.decompile(this.getInput().getAsFile().get(), this.getOutput().getAsFile().get(), options, libraries);
     }
 
     @Internal
     public AbstractDecompiler getAbstractDecompiler() {
```
lukebemish: Similar to the EnigmaProfile stuff above -- this is not conf cache compatible, and looks like a perfect place to use a build service.

buildSrc/src/main/java/quilt/internal/tasks/setup/DownloadMinecraftJarsTask.java
```diff
@@ -7,85 +7,93 @@
...
-public class DownloadMinecraftJarsTask extends DefaultMappingsTask {
+public abstract class DownloadMinecraftJarsTask extends DefaultMappingsTask {
     public static final String TASK_NAME = "downloadMinecraftJars";
-    @OutputFile
-    private final File clientJar;
+
+    @InputFile
+    public abstract RegularFileProperty getVersionFile();
+
+    @Internal("Fingerprinting is handled by getVersionFile")
```
lukebemish: Once again, remember that conf cache is a thing -- if it's used at task run time, use something serializable instead; if it's not, then making it a transient field _should_ work.

buildSrc/src/main/java/quilt/internal/tasks/setup/DownloadMinecraftLibrariesTask.java
```diff
@@ -3,88 +3,143 @@
...
+     * It should only be accessed from other {@linkplain  TaskAction tasks' actions} and via
+     * {@linkplain MapProperty lazy} {@linkplain org.gradle.api.tasks.Input input}.
+     */
+    public final Provider<Map<String, File>> artifactsByUrl;
+
     @InputFile
-    private final RegularFileProperty versionFile;
+    public abstract RegularFileProperty getVersionFile();
+
+    @Internal("Fingerprinting is handled by getVersionFile")
```
lukebemish: Same as above

buildSrc/src/main/java/quilt/internal/tasks/setup/DownloadMinecraftLibrariesTask.java
```diff
@@ -3,88 +3,143 @@
...
             } catch (NoSuchElementException e) {
                 new RuntimeException("Unable to find artifact for " + library.getName(), e).printStackTrace();
             }
 
             synchronized (lock) {
-                getProject().getDependencies().add("decompileClasspath", library.getName());
+                // TODO this is screwy.
+                //  Could we put these in a configuration in MappingsPlugin
+                //  and pass the configuration to an input of this task?
+                this.getProject().getDependencies().add("decompileClasspath", library.getName());
```
lukebemish: So, basically -- you cannot, and should not, add dependencies at task execution time! You can expect all dependencies to be set up before then and need to set them up before then. You also should not pass around `Configuration`s as task inputs/properties used at task run time (config cache incompatible if nothing else) If you need this info at that time to do so -- perhaps a `ValueSource` could do it instead of a task? Or, if having it be a task is still easiest, have them saved in a file that's generated by a task ran when _`buildSrc`_ is built, so that they're available in the main project?

buildSrc/src/main/java/quilt/internal/tasks/setup/DownloadWantedVersionManifestTask.java
```diff
@@ -2,59 +2,84 @@
...
+public abstract class DownloadWantedVersionManifestTask extends DefaultMappingsTask {
     public static final String TASK_NAME = "downloadWantedVersionManifest";
 
-    private final File manifestFile;
-    private final Optional<VersionEntry> manifestVersion;
+    @InputFile
+    public abstract RegularFileProperty getManifest();
+
+    @Internal("Fingerprinting is handled by getManifest and getReleaseTime")
+    protected abstract Property<VersionEntry> getManifestVersion();
```
lukebemish: Same worry as above.

buildSrc/src/main/java/quilt/internal/tasks/setup/ExtractServerJarTask.java
```diff
@@ -1,35 +1,37 @@
...
     public void extractServerJar() throws IOException {
-        FileUtils.copyFile(getProject().zipTree(this.getTaskByType(DownloadMinecraftJarsTask.class).getServerBootstrapJar()).matching(patternFilterable -> patternFilterable.include("META-INF/versions/*/server-*.jar")).getSingleFile(), serverJar);
-    }
-
-    public File getServerJar() {
-        return serverJar;
+        FileUtils.copyFile(
+            // TODO eliminate project access in task action
+            this.getProject()
+                .zipTree(this.getServerBootstrapJar().get())
```
lukebemish: Ah yes, our good friend `zipTree` that for some reason is not available on any non-internal injectable service -- this sucks unfortunately, and there's no nice solution other than building your own zip utility or splitting the task. Or possibly doing it in the task ctor? No clue if that laziness is lazy enough to work fine.

buildSrc/src/main/java/quilt/internal/tasks/setup/DownloadMinecraftLibrariesTask.java
```diff
@@ -3,88 +3,143 @@
...
-    private final RegularFileProperty versionFile;
+    public abstract RegularFileProperty getVersionFile();
+
+    @Internal("Fingerprinting is handled by getVersionFile")
+    protected abstract Property<Version> getVersion();
+
+    @OutputDirectory
+    public abstract DirectoryProperty getLibrariesDir();
+
+    @Internal("Fingerprinting is handled by getLibrariesDir")
```
lukebemish: Same as above

buildSrc/src/main/java/quilt/internal/tasks/unpick/gen/OpenGlConstantUnpickGeneratorTask.java
```diff
@@ -0,0 +1,389 @@
...
+        Pattern.compile("^org\\.lwjgl:lwjgl-opengl:([^:]*)$").asMatchPredicate();
+
+    @InputFile
+    public abstract RegularFileProperty getVersionFile();
+
+    @InputFile
+    public abstract RegularFileProperty getPerVersionMappingsJar();
+
+    @Input
+    public abstract MapProperty<String, File> getArtifactsByUrl();
```
lukebemish: I have a sneaking suspicion this is meant to be `@InputFiles`, if you're using the artifacts' contents.

buildSrc/src/main/java/quilt/internal/tasks/diff/CheckTargetVersionExistsTask.java
```diff
@@ -9,50 +9,54 @@
...
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
 import quilt.internal.Constants;
 import quilt.internal.tasks.DefaultMappingsTask;
 
-public class CheckTargetVersionExistsTask extends DefaultMappingsTask {
+public abstract class CheckTargetVersionExistsTask extends DefaultMappingsTask {
     public static final String TASK_NAME = "checkTargetVersionExists";
 
     @Internal
```
lukebemish: Uh... I don't like whatever this is doing. Does it set the field _during task execution_ then expect tasks running after it to grab the field from the task and use it? Iffy, to say the least, though I suppose functional. Regardless -- I'd annotate the method, not the field, here in either case.

buildSrc/src/main/java/quilt/internal/tasks/diff/CheckUnpickVersionsMatchTask.java
```diff
@@ -12,11 +12,11 @@
...
-public class CheckUnpickVersionsMatchTask extends DefaultMappingsTask {
+public abstract class CheckUnpickVersionsMatchTask extends DefaultMappingsTask {
     public static final String TASK_NAME = "checkUnpickVersionsMatch";
 
     @InputFile
-    private final RegularFileProperty unpickJson;
+    public abstract RegularFileProperty getUnpickJson();
 
     @Internal
     private boolean match = false;
```
lukebemish: Same as my worry with targetVersion above -- I'm not quite sold on this approach, it seems like a _bit_ of a bad pattern.

buildSrc/src/main/java/quilt/internal/tasks/setup/CheckIntermediaryMappingsTask.java
```diff
@@ -9,7 +9,10 @@
...
 import java.util.Set;
 
-public class CheckIntermediaryMappingsTask extends DefaultMappingsTask {
+// TODO eliminate this task (or make it a super class of tasks that currently depend on it)
+//  Tasks that depend on it should instead take the intermediary configuration as input and check if it exists and
+//  can be resolved
+public abstract class CheckIntermediaryMappingsTask extends DefaultMappingsTask {
     public static final String TASK_NAME = "checkIntermediaryMappings";
 
     @Internal
```
lukebemish: Ooh, here's another of whatever-the-heck these are!

lukebemish (commented):
Looks pretty good! Unsure if the eventual aim is config cache here (and maybe it ought to be, as config cache will be on by default in gradle 9 which is on the horizon) -- regardless, if so there's a bit of work to do, since a lot of stuff is still kinda unidiomatic in that regard; all things told though a _drastic_ improvement. Didn't comment on any remaining TODOs.

buildSrc/src/main/java/quilt/internal/tasks/setup/ExtractServerJarTask.java
supersaiyansubtlety: Yeah I've updated another task to save the `zipTree` in the constructor and it's working fine

supersaiyansubtlety: The issue I asked about on discord was because the other task was also *outputting* the file that was getting unzipped, so I had to separate the unzipping to another task.

supersaiyansubtlety:
~~Note: Rather than marking this PR as ready for review, when it's ready I'll be closing it and creating a new one.~~
~~That will clean up the messy history here on GitHub caused by rewriting history.~~
Never mind, it cleaned itself up, I underestimated GitHub.

buildSrc/src/main/java/quilt/internal/util/DownloadUtil.java
```diff
@@ -0,0 +1,98 @@
...
+     * @throws IOException see {@link FileUtils#copyURLToFile}
+     */
+    public static boolean downloadUnhandled(URL url, File dest, boolean overwrite, @Nullable Logger logger) throws
+            IOException {
+        if (overwrite || !dest.exists()) {
+            if (logger != null) {
+                logger.lifecycle("Downloading {}", url);
+            }
+
+            // TODO QUESTION I just made these timeouts up, they could probably be better
```
supersaiyansubtlety: Suggestions?

buildSrc/src/main/java/quilt/internal/task/decompile/DecompileVineflowerTask.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    @TaskAction
+    public void decompile() throws IOException {
+        FileUtils.deleteDirectory(this.getOutput().get().getAsFile());
+
+        // TODO QUESTION this sometimes logs a non-fatal exception:
+        //  java.nio.file.FileSystemAlreadyExistsException
+        //  it's thrown from
+        //  org.jetbrains.java.decompiler.main.plugins.JarPluginLoader:init
+        //  it seems harmless
```
supersaiyansubtlety: Please verify this is harmless

buildSrc/src/main/java/quilt/internal/task/ArtifactFileTask.java
```diff
@@ -0,0 +1,163 @@
...
+import org.gradle.api.tasks.Internal;
+import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.OutputFile;
+
+// TODO QUESTION Is there a way to make it so MavenPublication#artifact will accept
+//  these tasks directly and run them to build their outputs, similar to AbstractArchiveTask?
+//  (eliminating the need for the artifact convenience methods)
+//  I considered implementing PublishArtifact on this, but it's annotated with @HasInternalProtocol.
+//  This suggests that it would work but doesn't recommend it:
+//  https://github.com/gradle/gradle/issues/17273#issuecomment-858400396
```
supersaiyansubtlety: Suggestions?

buildSrc/src/main/java/quilt/internal/plugin/MapV2Plugin.java
```diff
@@ -0,0 +1,261 @@
...
+            mapMinecraftJarsTasks.insertAutoGeneratedMappings();
+
+        // register this plugin's tasks
+        final TaskContainer tasks = project.getTasks();
+
+        final var mergeTinyV2 = tasks.register(
+            MergeTinyV2Task.MERGE_TINY_V_2_TASK_NAME,
+            MergeTinyV2Task.class,
+            task -> {
+                // TODO QUESTION this used to be dependent on v2UnmergedMappingsJar, but afaict it has no effect on this task
```
supersaiyansubtlety: Please check this

supersaiyansubtlety (commented):
A few remaining questions:

resolved: probably not much benefit, but maybe worth doing in a followup PR
~~Should I just throw `@CachableTask` on everything?~~
~~Do we benefit much from build cache without it?~~

resolved: seems fine
~~Please check the last several commits where I messed with the caching of a few tasks.~~

Also please manually test this into oblivion.

all resolved
~~TODO QUESTIONS:~~

build.gradle
```diff
@@ -1,403 +1,178 @@
...
-	def docletResources = sourceSets.doclet.resources.asFileTree
-
-	failOnError = true // Failing on error is important to ensure that a Javadoc jar is available.
+	// Failing on error is important to ensure that a Javadoc jar is available.
+	failOnError = true
 	maxMemory = '2G'
 
-	// verbose = true // enable to debug
+	// enable to debug
+	final debug = false
```
OroArmor: Can we make this a property? maybe something like `-Dquilt.mappings.debug.javadoc`?

buildSrc/src/main/java/quilt/internal/constants/Namespaces.java
```diff
@@ -0,0 +1,8 @@
+package quilt.internal.constants;
+
+public interface Namespaces {
+    String INTERMEDIARY = Constants.INTERMEDIARY_MAPPINGS_NAME;
+    String PER_VERSION = Constants.PER_VERSION_MAPPINGS_NAME;
```
OroArmor: I would make this the source of truth and have the others pull from this. Also I would move intermediary to the bottom. We might also want a rename of `per_version` to `intermediate` (not intermediary, thanks fabric), since the idea was to make the QM buildscript mostly independent of the middle layer for the mappings

buildSrc/src/main/java/quilt/internal/plugin/MapV2Plugin.java
OroArmor: Yarn doesnt have `v2UnmergedMappingsJar` anymore, its probably fine

buildSrc/src/main/java/quilt/internal/task/EnigmaProfileConsumingTask.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Holds any {@value org.quiltmc.enigma_plugin.Arguments#SIMPLE_TYPE_FIELD_NAMES_PATH}s
+     * configuration files obtained from {@link #getEnigmaProfileService() enigmaProfileService}'s
+     * {@link EnigmaProfileService#getProfile() profile}.
+     * <p>
+     * {@link EnigmaProfileConsumingTask}s may not access these files directly, but they affect Enigma's behavior,
+     * so they must be considered for up-to-date checks.
+     */
+    @InputFiles
+    ConfigurableFileCollection getSimpleTypeFieldNamesFiles();
```
OroArmor: Could we change this to something like `getProfileFileDependencies` or something? i have a plugin PR that adds another file to this list and this name is somewhat confusing

buildSrc/src/main/java/quilt/internal/task/MappingsTask.java
```diff
@@ -0,0 +1,9 @@
+package quilt.internal.task;
+
+import org.gradle.api.Task;
+import quilt.internal.plugin.QuiltMappingsPlugin;
+
+/**
+ * All tasks added by {@link QuiltMappingsPlugin QuiltMappingsPlugin} share this type.
+ */
+public interface MappingsTask extends Task { }
```
OroArmor: Do we need this anymore or at all?

unpick-definitions/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Put the auto generated files here
```
OroArmor, line 1: Were these moved somewhere else?

OroArmor (changes requested):
Few nitpicks, WAY larger than i expected, but looks *amazing*. great job!

supersaiyansubtlety:
resolved
~~Forgot one question:~~

~~These properties should be versions in the catalog, but `gson` and `netty` already have different, unused versions in the catalog:~~
- ~~https://github.com/supersaiyansubtlety/quilt-mappings/blob/a1c4c484c3aa329859f59ca6961ac40700eafad8/gradle.properties#L8-L12~~
- ~~https://github.com/supersaiyansubtlety/quilt-mappings/blob/a1c4c484c3aa329859f59ca6961ac40700eafad8/gradle/libs.versions.toml#L22-L23~~

~~Which versions of `gson` and `netty` do we want?~~

buildSrc/src/main/java/quilt/internal/task/MappingsTask.java
supersaiyansubtlety: Not really, I kept it in case we wanted to add commonly injected services or broad convenience methods similar to `MappingsProjectPlugin`, or in case we want to identify plugin tasks by type.

But since the PR is basically ready and this isn't used I think that's unlikely.

If we remove it we can also remove `DefaultMappingsTask` and just set groups in each task.

unpick-definitions/.gitignore
supersaiyansubtlety: `build/mappings/`:
https://github.com/supersaiyansubtlety/quilt-mappings/blob/a1c4c484c3aa329859f59ca6961ac40700eafad8/buildSrc/src/main/java/quilt/internal/plugin/MapV2Plugin.java#L130-L136

supersaiyansubtlety:
Re: [Configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html)

I tried running some tasks with `--configuration-cache` and got:
> Configuration cache state could not be cached: field `pendingMap` of `org.gradle.api.internal.DefaultNamedDomainObjectCollection$UnfilteredIndex` bean found in field `delegate` of `org.gradle.api.internal.DefaultNamedDomainObjectCollection$FilteredIndex` bean found in field `index` of `org.gradle.api.internal.tasks.DefaultTaskCollection` bean found in field `delegate` of `org.gradle.api.internal.tasks.DefaultRealizableTaskCollection` bean found in input property `$1` of task `:buildMappingsTiny` of type `quilt.internal.task.build.BuildMappingsTinyTask`: error writing value of type 'java.util.LinkedHashMap'
> java.util.ConcurrentModificationException (no error message)

That's for `buildMappingsTiny`, I got a similar error mentioning `error writing value of type 'java.util.LinkedHashMap'` for `findDuplicateMappingFiles`.

idk how to fix these because idk where the `LinkedHashMap` comes from.

configuration caching will have to go in a followup PR
~~If it turns out it wouldn't take too many changes to fix config caching I'm ok with including it in this PR, otherwise it'll have to go in a followup PR.~~

eristhea:
is there anything that remains to be done on this now that #659 is merged?

supersaiyansubtlety:
> is there anything that remains to be done on this now that #659 is merged?

~~It's not waiting on any other PRs, but I need to merge changes and make sure tests work.~~ done

~~There are also a few unresolved questions above.~~ resolved

buildSrc/src/main/java/quilt/internal/task/MappingsTask.java
supersaiyansubtlety: eliminated both

OroArmor (approved):
Honestly too big to properly review now. From my previous looks its going great. I trust that you did it correctly

buildSrc/src/main/java/quilt/internal/util/DownloadUtil.java
supersaiyansubtlety: these have worked so far, we can lengthen them later if needed

buildSrc/src/main/java/quilt/internal/task/decompile/DecompileVineflowerTask.java
supersaiyansubtlety: hasn't cause any issues afaict
I've left the comment but remove the TODO

PR #646, opened 2024-10-06
Named methods in `CreeperEntity` and `World`.

HyperPigeon:
- Mapped public boolean method in `CreeperEntity` to `isCharged`.
- Mapped multiple public void methods in `World` to `createExplosion`.


supersaiyansubtlety (approved):

supersaiyansubtlety:
Actually, maybe`spawnExplosion` instead?
`createExplosion` sounds like it would return an `Explosion` instance.

HyperPigeon:
> Actually, maybe`spawnExplosion` instead? `createExplosion` sounds like it would return an `Explosion` instance.

That makes a lot of sense. I'll update the mappings.

supersaiyansubtlety (approved):

OroArmor (approved):

PR #653, opened 2024-10-11
rename LivingEntity setCurrentHand -> tryStartUsing and add javadoc 

supersaiyansubtlety:
Setting the current hand is only one of the things it does:
```java
/**
 * If the passed {@link net.minecraft.util.Hand hand} holds an
 * {@link net.minecraft.item.ItemStack item} and one is not already being used,
 * starts using the held {@link net.minecraft.item.ItemStack item}.
 */
public void tryStartUsing(Hand hand) {
    ItemStack lv = this.getStackInHand(hand);
    if (!lv.isEmpty() && !this.isUsingItem()) {
        this.activeItemStack = lv;
        this.itemUseTimeLeft = lv.getUseTicks(this);
        if (!this.getWorld().isClient) {
            this.setLivingFlag(1, true);
            this.setLivingFlag(2, hand == Hand.OFF_HAND);
            this.emitGameEvent(GameEvent.ITEM_INTERACT_START);
        }
    }
}
```


mappings/net/minecraft/entity/LivingEntity.mapping
```diff
@@ -120,7 +120,7 @@ CLASS net/minecraft/unmapped/C_usxaxydn net/minecraft/entity/LivingEntity
 	METHOD m_awmtskvv getMovementSpeed (F)F
 		ARG 1 slipperiness
 	METHOD m_axkhmdzu getDeathSound ()Lnet/minecraft/unmapped/C_avavozay;
-	METHOD m_ayeajhdm setCurrentHand (Lnet/minecraft/unmapped/C_laxmzoqs;)V
+	METHOD m_ayeajhdm startUsing (Lnet/minecraft/unmapped/C_laxmzoqs;)V
```
eristhea:
```suggestion
	METHOD m_ayeajhdm tryStartUsing (Lnet/minecraft/unmapped/C_laxmzoqs;)V
```
i would prefix with `try`, since it won't start using unless there's a valid item

eristhea (changes requested):

mappings/net/minecraft/entity/LivingEntity.mapping
supersaiyansubtlety: good point, I'll add a javadoc, too

eristhea (approved):
gorge!

IotaBread (approved):

PR #654, opened 2024-10-13
improve movement terminology

supersaiyansubtlety:
rename "movement speed" things -> "acceleration factor"

They're used to calculate the vector added to velocity:
```java
// Entity
public void updateVelocity(float accelerationFactor, Vec3d movementInput) {
    Vec3d lv = movementInputToAcceleration(movementInput, accelerationFactor, this.m_ndosmusf());
    this.setVelocity(this.getVelocity().add(lv));
}
```

Also, it didn't make sense for higher `slipperiness` to reduce the result of this method (formerly named `getMovementSpeed`):
```java
// LivingEntity
private float getAccelerationFactor(float slipperiness) {
    return this.isOnGround() ? this.getBaseAccelerationFactor() * (0.21600002F / (slipperiness * slipperiness * slipperiness)) : this.getAirAccelerationFactor();
}
```
... but it *does* make sense for `slipperiness` to reduce acceleration.


eristhea (approved):
looks great!

OroArmor (approved):

PR #666, opened 2024-11-22
Dispenser stuff

supersaiyansubtlety:
- completes `n/m/block/dispenser/` and `CauldronBehavior`
- renames `ItemDispenserBehavior` -> `DefaultDispenserBehavior`
- rename `FallibleItemDispenserBehavior` -> `FallibleDispenserBehavior`
- renames  `n/m/util/math/BlockPointer` -> `n/m/block/dispenser/DispenserContext`:
it holds a `DispenserBlockEntity` and is used by `DispenserBehavior`s, context is more conventional than pointer
- consistently names the methods called in `Bootstrap::bootstrap` `bootstrap`


mappings/net/minecraft/Bootstrapper.mapping
```diff
@@ -1,5 +1,5 @@
-CLASS net/minecraft/unmapped/C_ntskruzf net/minecraft/Bootstrap
-	FIELD f_jltdqkzo initialized Z
+CLASS net/minecraft/unmapped/C_ntskruzf net/minecraft/Bootstrapper
```
eristhea: i disagree with this rename, since it's an all-static class. additionally, `Bootstrap` has parity across yarn and moj

eristhea (changes requested):
looks great, just one issue!

mappings/net/minecraft/Bootstrapper.mapping
supersaiyansubtlety: I mainly did it to avoid `Bootstrap.bootstrap()`, bit ig that's fine.

Do you prefer the old "initialize" field+method names as well?

eristhea: i think the new ones are better!

src/packageDocs/java/net/minecraft/package-info.java
```diff
@@ -7,7 +7,7 @@
  * in their intermediary names.
  *
  * <p>Unmapped classes go into this package by default. This package additionally contains
- * {@link Bootstrap}, {@link SharedConstants}, and {@link MinecraftVersion} classes.
+ * {@link com.sun.jdi.Bootstrap}, {@link SharedConstants}, and {@link MinecraftVersion} classes.
```
supersaiyansubtlety: oops, fixing

eristhea (approved):

mappings/net/minecraft/world/WorldEvents.mapping
IotaBread, line 1: Javadocs in this class still use `n.m.util.math.BlockPointer` instead of `n.m.block.dispenser.DispenserContext`

supersaiyansubtlety: fixed

IotaBread (approved):

PR #667, opened 2024-11-23
Miscellaneous refactorings

### LINKED ISSUE #484: `AbstractClientPlayerEntity#getSpeed` is poorly named
It has to do with FOV modification, and mojmap is `getFieldOfViewModifier`. Can 100% have a better name.

### LINKED ISSUE #662: SpriteButtonWidget#Builder variable/constructor parameter `displayMessage` has its meaning reversed (1.21.1)
- The message is displayed when set to `false` and hidden when set to `true`

A better name would probably be something along the lines of `spriteOnly` as this would also reflect the class name of the widget used if set to true: `SpriteOnlyButtonWidget`.

maloryware:
Uses the same name as Yarn mappings. 

Could be worth considering whether it's worth specifying that the multiplier primarily takes the player speed into account, but as far as I can tell, nothing else changes the field of view besides speed-altering effects.

Closes #484

### DIFF
--- a/mappings/net/minecraft/client/network/AbstractClientPlayerEntity.mapping
+++ b/mappings/net/minecraft/client/network/AbstractClientPlayerEntity.mapping
@@ -9,6 +9,6 @@ CLASS net/minecraft/unmapped/C_sgyyxgui net/minecraft/client/network/AbstractCli
 		ARG 2 profile
 	METHOD m_ipxgqpif lerpVelocity (F)Lnet/minecraft/unmapped/C_vgpupfxx;
 		ARG 1 delta
-	METHOD m_jjnbolzs getSpeed (ZF)F
+	METHOD m_jjnbolzs getFovMultiplier (ZF)F
 	METHOD m_kzmhnjvr getSkin ()Lnet/minecraft/unmapped/C_idvschhb;
 	METHOD m_luyaebra getPlayerListEntry ()Lnet/minecraft/unmapped/C_rdnfmxue;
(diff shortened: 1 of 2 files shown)


mappings/net/minecraft/client/network/AbstractClientPlayerEntity.mapping
```diff
@@ -9,6 +9,6 @@ CLASS net/minecraft/unmapped/C_sgyyxgui net/minecraft/client/network/AbstractCli
 		ARG 2 profile
 	METHOD m_ipxgqpif lerpVelocity (F)Lnet/minecraft/unmapped/C_vgpupfxx;
 		ARG 1 delta
-	METHOD m_jjnbolzs getSpeed (ZF)F
+	METHOD m_jjnbolzs getFovMultiplier (ZF)F
```
supersaiyansubtlety, line 12:
```suggestion
	METHOD m_jjnbolzs getFovMultiplier (ZF)F
		ARG 1 firstPerson
		ARG 2 effectScale
```
might as well name the params too

supersaiyansubtlety (approved):

maloryware (commented):
Is it worth replacing `Message` with `Text`? Unsure if 'message' comes off as clear enough

supersaiyansubtlety:
> Is it worth replacing `Message` with `Text`? Unsure if 'message' comes off as clear enough

I think message is good

eristhea (approved):
looks good!

PR #669, opened 2024-11-27
fix DownloadTargetMetaFileTask when meta file is unavailable

supersaiyansubtlety:
fixes <https://github.com/QuiltMC/quilt-mappings/actions/runs/12044298644/job/33581079167#step:6:595>

Note: This is just a quick fix, I'll make a PR that improves upon it:
- `DownloadTargetMetaFileTask` could be converted to a `ValueSource`
- we could maybe ping the url instead of just trying to download and catching the exception if it fails


supersaiyansubtlety:
Superseded by #670

PR #672, opened 2024-12-04
fix 'overrides final method' error

supersaiyansubtlety:
Fixes this error which occurs on launch when using 1.21.4-build.1:
```
java.lang.IncompatibleClassChangeError: class net.minecraft.entity.InteractionEntity overrides final method net.minecraft.entity.Entity.getHeight()F
```


eristhea (commented):
should the corresponding fields (f_luxekrhz and f_rmbogufh) be updated as well?

supersaiyansubtlety:
> should the corresponding fields (f_luxekrhz and f_rmbogufh) be updated as well?

arguably, but their keys are just `"width"` and `"height"`

I think it's fine either way, what's your preference?

eristhea:
personally, we should update the field names but not the key fields

eristhea (approved):
merging after checks!

PR #686, opened 2025-01-06
1.21.4 port

supersaiyansubtlety:
ports 1.21.1 changes to 1.21.4, including updating engima to 2.5.3
also restores `GlStateManager` from 1.21.3 which was somehow missing on 1.21.4


eristhea (approved):
thanks!

PR #687, opened 2025-02-16
Mappings for QSL update

supersaiyansubtlety:
- restores `MapColor`
  - refactors `getPackedId` -> `packIds`: brief and it packs two ids (map color and  brightness)
  - everything else is copied from 1.21.3
- restores several `AbstractFurnaceBlockEntity` fields
- maps `EntityRenterState` and `LivingEntityRenderState`
- mostly completes `EntityRenderer` and `LivingEntityRenderer`
- mostly completes `GuiGraphics`
  - renames `color` params -> `argb`
  - some param names are from yarn


Pixaurora:
Edit: Added more stuff that's not mapped. Thanks for helping complete the mappings!

* classes similar to  `MobEntityRenderer` now have a "render state" generic class attached which can store some data, such as if a skeleton is holding a bow.
* `drawGuiTexture` and `drawTexture` methods of `GuiGraphics` now take a `RenderLayer` argument, and they became unmapped.

supersaiyansubtlety:
I'll get those mapped!

mappings/net/minecraft/client/render/entity/LivingEntityRenderState.mapping
```diff
@@ -0,0 +1,25 @@
+CLASS net/minecraft/unmapped/C_ozphnurq net/minecraft/client/render/entity/LivingEntityRenderState
+	FIELD f_axhqulem pose Lnet/minecraft/unmapped/C_ufdjspmk;
+	FIELD f_cyjcipbw customName Lnet/minecraft/unmapped/C_rdaqiwdt;
+	FIELD f_fovngtjq skullType Lnet/minecraft/unmapped/C_hxsyxqpk$C_nyiksgwh;
+	FIELD f_hdaqnjbp headBodyYawDiff F
+	FIELD f_ihpohsqn isHurt Z
+	FIELD f_immbtgmx invisibleToPlayer Z
+	FIELD f_kcknwmag wornHeadAnimationProgress F
```
eristhea, line 8:
```suggestion
	FIELD f_kcknwmag headFeatureAnimationProgress F
```
maybe? the name is a little confusing, i'm not sure if this is better

mappings/net/minecraft/client/render/item/ItemRenderer.mapping
```diff
@@ -43,3 +55,6 @@ CLASS net/minecraft/unmapped/C_xjqetogb net/minecraft/client/render/item/ItemRen
 		ARG 0 provider
 		ARG 1 layer
 		ARG 2 entry
+	METHOD m_ymofutkb getColor ([II)I
+		ARG 0 colors
+		ARG 1 i
```
eristhea:
```suggestion
		ARG 1 index
```
i don't really like `i` for `index` in general, it's a little ambiguous

mappings/net/minecraft/block/MapColor.mapping
```diff
@@ -0,0 +1,101 @@
+CLASS net/minecraft/unmapped/C_vhpmxtfv net/minecraft/block/MapColor
+	COMMENT Represents the surface color of a block when rendered from the {@link net.minecraft.client.render.MapRenderer}.
```
eristhea: i'd link to the instance in `MinecraftClient` -- the usage of `the` makes it seem like that's the intent

eristhea (commented):
looks great!

mappings/net/minecraft/client/render/entity/LivingEntityRenderState.mapping
supersaiyansubtlety: It's specifically for when they're wearing an `AbstractSkullBlock` on their head, so I think it should include `worn`.
Though maybe `head` -> `skull`?

I kind of want to rename `AbstractSkullBlock` -> `AbstractHeadBlock`, but its block entity uses the string `"skull"` : /

eristhea: i think this name is a fine compromise, i don't think skull clarifies much

eristhea (approved):

PR #688, opened 2025-02-20
Gametest mappings

supersaiyansubtlety:
Adds `net.minecraft.test` coverage, improves names, and adds javadocs.

- completes `TestContext` with minor refactors and adds javadocs
- renames and completes `TimedTaskRunner` -> `TestTaskSequence`
  - substantial refactors including `createAndAdd` -> `await` and `execute...` -> `then...`
  - adds javadocs
- renames `TestUtil` -> `TestsRunner`
- maps some of the new snapshot gametest classes
- maps miscellaneous tangential bits


mappings/net/minecraft/block/entity/BoundsRenderable.mapping
```diff
@@ -0,0 +1,12 @@
+CLASS net/minecraft/unmapped/C_qdasistl net/minecraft/block/entity/BoundsRenderable
```
eristhea: is there a reason this isn't in natural word order?

mappings/net/minecraft/test/TestTaskSequence.mapping
```diff
@@ -0,0 +1,124 @@
...
+	METHOD m_jtywgcow runOrFail (Ljava/lang/Runnable;)V
+		ARG 1 task
+	METHOD m_kxztjfwm succeed ()V
+		COMMENT Causes the {@link #test} to succeed once prior added tasks have completed.
+	METHOD m_ndryulqk await (Ljava/lang/Runnable;)Lnet/minecraft/unmapped/C_slbbddbd;
+		COMMENT Once reached, the passed {@code task} will repeat until it succeeds or the
+		COMMENT {@link #test} times out.
+		ARG 1 task
+	METHOD m_negtjhjf fail (Ljava/util/function/Supplier;)V
+		COMMENT Cuases the {@link #test} to fail with an exception created by the passed
```
eristhea:
```suggestion
		COMMENT Causes the {@link #test} to fail with an exception created by the passed
```
spellcheck where were you

eristhea (approved):
amazing! these are all so high quality and i love the mountains of javadoc

mappings/net/minecraft/block/entity/BoundsRenderable.mapping
supersaiyansubtlety: The idea was `Renderable` was the noun, but it's awkward.
I don't particularly like `RenderableBounds` because it seems off to me to call a `StructureBlockEntity` or a `TestInstanceBlockEntity` "bounds"; the bounds aren't primarily what they *are*, they're just something they have.

How about `BoundedRenderable`?

eristhea: i like that!

PR #689, opened 2025-03-09
Misc fixes and mappings

supersaiyansubtlety:
Miscellaneous fixes and tweaks, along with some new things I got sidetracked on.

Notably:
 - renames `n.m.data...Model[s]` -> `ModelTemplate[s]`: it's for generating children of parent template models
 - renames all `...Entity::create[X]Attributes` -> `buildAttributes`: consistent and hints that subclasses should build off their superclass' method


mappings/net/minecraft/entity/vehicle/TntMinecartEntity.mapping
```diff
@@ -1,5 +1,10 @@
 CLASS net/minecraft/unmapped/C_joafvsgg net/minecraft/entity/vehicle/TntMinecartEntity
+	FIELD f_fiseepir EXPLOSION_SPEED_FACTOR_KEY Ljava/lang/String;
 	FIELD f_fuwpndte PRIME_TNT_STATUS B
+	FIELD f_lpqzbazb explosionSpeedFactor F
+	FIELD f_tklralig DEFAULT_EXPLOSION_SPEED_FACTOR F
+	FIELD f_uhppptgz FUZE_KEY Ljava/lang/String;
```
eristhea:
```suggestion
	FIELD f_uhppptgz FUSE_KEY Ljava/lang/String;
```
i assume?

eristhea (approved):
looks great, love the refactors! `buildAttributes` especially is a great change

mappings/net/minecraft/entity/vehicle/TntMinecartEntity.mapping
supersaiyansubtlety: yep, good catch

PR #690, opened 2025-03-21
Port #687

supersaiyansubtlety:
Ports #687 to 1.21.5-rc1

Should not be squash-merged for the sake of making future ports cleaner.


eristhea (approved):
gorgeous!

PR #692, opened 2025-03-24
gamtest more

supersaiyansubtlety:
Completes n.m.test


mappings/net/minecraft/test/TestBatchUtil.mapping
```diff
@@ -0,0 +1,33 @@
+CLASS net/minecraft/unmapped/C_doimveny net/minecraft/test/TestBatchUtil
```
eristhea: really minor change, but i think `TestBatchingUtil` has a better sound. up to you!

eristhea (approved):
looks super clean! very few comments from me today :p

PR #697, opened 2025-04-04
Gametest verification

supersaiyansubtlety:
Adds a single gametest using the local mappings produced by the mappings project.

Adds a step to the Build workflow that runs `build runGametest` in the gametest directory.

Meant to catch things like #695 and #696.
Building also catches things like #698.

The gametest:
- (implicitly) verifies bootstrap bootstraps
- (implicitly) loads a world
- spawns and kills each entity type

~~I could add a non-game test to explicitly verify bootstrap bootstraps so it fails earlier, but I'm not sure if it's worth it.~~
Can't do that without fapi.

~~This moves basically the entire mappings project into an includeBuild'ed `core` project.~~

~~As a result, all the mappings project tasks must be prefixed with `:core:`, and actions will need to be updated.~~

Thanks to @lukebemish for help with getting loom to use local mappings!


lukebemish:
Is there a reason the root project was moved to a subdirectory? Why not keep it where it is and have the game tests in a subdirectory?

supersaiyansubtlety:
> Is there a reason the root project was moved to a subdirectory? Why not keep it where it is and have the game tests in a subdirectory?

Since the gametest project includeBuild’s the core project, I’m not sure how I could keep the core project in the root directory without creating circular project dependencies. 
I’m also not sure how I’d refer to it for includeBuild, does it accept relative paths like ‘..’?

lukebemish:
Core project in root project, with no dependency on the gametest project at all. Gametest project includebuilds it's parent directory: `includeBuild '../'`.

supersaiyansubtlety:
If the gametest project is neither `include`d nor `includeBuild`ed, how is it recognized by gradle at all?
Or are you saying we change the working directory to the gametest project before running the gametest task?

lukebemish:
Just run the task from that directory, yes. GH actions have a working directory option -- you can reference `../gradlew` just fine within there. There's really no good reason for that test to be part of the main build given how this all works.

build.gradle
```diff
@@ -324,6 +367,13 @@ publishing {
 
 			// See below the publishing block for intermediary artifact publishing
 		}
+
+		// for gametest project
+		gametestable(MavenPublication) {
+			groupId 'org.quiltmc'
+			artifactId 'gametestable-quilt-mappings'
+			from components.gametestable
+		}
```
lukebemish: Is there a reason the gametest project can't just consume the _normal_ publication published to that local maven? After all, it's all running in an earlier build step so you don't actually need the component/consumable variant stuff.

lukebemish (changes requested):

build.gradle
supersaiyansubtlety: Oh was the component only required for dependency substitution?
I'll try without it.

supersaiyansubtlety: Unless I made a mistake, it seems it needs the component publishing for gametest to resolve it
<https://github.com/supersaiyansubtlety/quilt-mappings/tree/gametest-verification-no-component-publishing>

>FAILURE: Build failed with an exception.
> 
> * What went wrong:
> A problem occurred configuring root project 'gametest'.
> > Could not resolve all dependencies for configuration ':mappings'.
   > Could not find org.quiltmc:gametestable-quilt-mappings:1.21.5+build.local.
     Searched in the following locations:
       ...
       - file:/C:/Users/super/Projects/Minecraft/Modding/Quilt/quilt-mappings/build/gametestable/org/quiltmc/gametestable-quilt-mappings/1.21.5+build.local/gametestable-quilt-mappings-1.21.5+build.local.pom
     Required by:
         root project :

supersaiyansubtlety: oh duh, I forgot to remove `gametestable-` from the dependency

gametest/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+        fabric_loader_version: libs.versions.fabric.loader.get(),
+        fabric_api_version: libs.versions.fabric.api.get()
+    ]
+
+    inputs.properties(expanding)
+
+    filesMatching('fabric.mod.json') {
+        expand expanding
+    }
+}
```
lukebemish: Trailing newline please? I know, minor thing, probably no big deal...

lukebemish (approved):
Gradle-wise, looks good to me!

eristhea:
just wanted to pop in and say this is INCREDIBLE i'm so amazed that you cooked this up

i'm still a bit too busy to really dig into reviewing, but i appreciate this work sm and i'll give it a look within a week or so!

eristhea (changes requested):
i would prefer that these versions were situated in the same file as all the other versions. that way we can more easily auto-update them in matcher!

also i had to remove the mc version constraint since it didn't work on snapshots (floader messes with the version number, making the snapshot `1.21.5-alpha.25.w.16.a` or something like that). i don't see a reason to fix the constraint so i just removed it

supersaiyansubtlety:
thanks for catching the snapshot issue!

supersaiyansubtlety:
got permissions sorted, so I've updated the build workflow with a step that runs `build runGametest` in the gametest directory

eristhea (approved):
now that i'm thinking of it, do we maybe want to query the FAPI version at runtime and cancel if no version exists?
this fixes the issue of FAPI not always being available to update during matching (which would mean we have to go in later for every snapshot and update the version), but it might introduce a lack of reproducibility

supersaiyansubtlety:
~~Hm, I wonder how doable it would be to have our own mixin-based entry point and use vanilla's data-driven stuff
That would avoid the issue entirely (and I could switch back to qloader).~~

~~I'll look into that soon.~~

Done!

gametest/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+    ]
+
+    inputs.properties(expanding)
+
+    filesMatching('quilt.mod.json') {
+        expand expanding
+    }
+}
+
+// mod test data isn't loaded without fapi, so copy it to a datapack
```
supersaiyansubtlety, line 60: Just pointing this out to make sure we're ok with this additional jank.

This is easier and hopefully more maintainable than a mixin to load it from resources.

supersaiyansubtlety:
Something in one of the recent snapshots changed things so the gametest server hangs, I'm looking into it.

supersaiyansubtlety:
> Something in one of the recent snapshots changed things so the gametest server hangs, I'm looking into it.

fixed

supersaiyansubtlety:
Note that currently, due to the way loom caches things, it's often necessary to use `--refresh-dependencies` when building or running the gametest locally. If you've previously built, then update mappings, those changes won't be reflected without `--refresh-dependencies`.

This isn't a problem for the build workflow because of containerization.

lukebemish:
I tracked down why that happens -- it's because the gametest build isn't pulling a new version of the relevant stuff from the dummy maven it's published to. That should be fixable by marking the dependency as changing -- literally just `changing = true` on the dependency.

gametest/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+        // do not cache mappings
+        cacheChangingModulesFor 0, SECONDS
+    }
+}
+
+dependencies {
+    minecraft(rootLibs.minecraft)
+    mappings ("org.quiltmc:quilt-mappings:${gametestable.version}:intermediary-v2") {
+        changing = true
+    }
```
supersaiyansubtlety, line 40: This unfortunately didn't fix it.

OroArmor (approved):
Looks good. Do we need the second copy of gradle?

supersaiyansubtlety:
> Do we need the second copy of gradle?

Nope, removed

OroArmor (approved):

Issue #665, opened 2024-11-18
Create conventions for `T get(K keyLike)` variants


supersaiyansubtlety:
It's been discussed a couple times in PR comments, I propose the following conventions:
```java
// net/minecraft/state/State

@Nullable
public <T extends Comparable<T>> T get(Property<T> property) {
    Comparable<?> comparable = (Comparable<?>)this.entries.get(property);
    return comparable == null ? null : property.getType().cast(comparable);
}

public <T extends Comparable<T>> T getOrThrow(Property<T> property) {
    Comparable<?> comparable = (Comparable<?>)this.entries.get(property);
    if (comparable == null) {
        throw new IllegalArgumentException("Cannot get property " + property + " as it does not exist in " + this.owner);
    } else {
        return property.getType().cast(comparable);
    }
}

public <T extends Comparable<T>> Optional<T> getOrEmpty(Property<T> property) {
    return Optional.ofNullable(this.get(property));
}

public <T extends Comparable<T>> T getOrDefault(Property<T> property, T defaultValue) {
    return Objects.requireNonNullElse(this.get(property), defaultValue);
}
```

I've seen `getOrNull` suggested for the first variant.
I prefer `get` because I think one of them should be just `get`, and this matches `java.util.Map::get`.

If only one single-param variant is present, I think `get` should be used regardless of which variant it is (`getOrDefault` should always be `getOrDefault`).

Note that I'm not suggesting a convention for differentiating between methods that get values based on various different key-likes, such as `Identifier id` and `int rawId`, which I don't think should be generalized into a convention.

PR #699, opened 2025-04-15
Entity work

supersaiyansubtlety:
`n.m.entity` coverage

also starts mapping `n.m.world.waypoint`

### DIFF
--- a/mappings/net/minecraft/component/type/AttributeModifiersComponent.mapping
+++ b/mappings/net/minecraft/component/type/AttributeModifiersComponent.mapping
@@ -8,10 +8,15 @@ CLASS net/minecraft/unmapped/C_azfkqhlm net/minecraft/component/type/AttributeMo
 	METHOD m_bpimhffk forEachModifier (Lnet/minecraft/unmapped/C_qfoqahef;Ljava/util/function/BiConsumer;)V
 		ARG 1 slots
 		ARG 2 consumer
+	METHOD m_btrvtwgo appendTooltip (Lnet/minecraft/unmapped/C_qfoqahef;Lorg/apache/commons/lang3/function/TriConsumer;)V
+		ARG 1 slots
+		ARG 2 appender
 	METHOD m_gdqixgmv with (Lnet/minecraft/unmapped/C_cjzoxshv;Lnet/minecraft/unmapped/C_hdbqsqsm;Lnet/minecraft/unmapped/C_qfoqahef;)Lnet/minecraft/unmapped/C_azfkqhlm;
 		ARG 1 attribute
 		ARG 2 modifier
 		ARG 3 group
+	METHOD m_mjzxuzjp (Ljava/text/DecimalFormat;)V
+		ARG 0 format
 	METHOD m_pxeavngw builder ()Lnet/minecraft/unmapped/C_azfkqhlm$C_vinsjuvn;
 	METHOD m_zgatrfmf forEachModifier (Lnet/minecraft/unmapped/C_yuycoehb;Ljava/util/function/BiConsumer;)V
 		ARG 1 slot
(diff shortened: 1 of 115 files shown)


mappings/net/minecraft/component/type/AttributeModifiersComponent.mapping
```diff
@@ -23,6 +28,37 @@ CLASS net/minecraft/unmapped/C_azfkqhlm net/minecraft/component/type/AttributeMo
...
+			ARG 3 attribute
+			ARG 4 modifier
+		METHOD m_oopxewmk (Lnet/minecraft/unmapped/C_azfkqhlm$C_kqgfvdat$C_ojlwbbdy;)Lcom/mojang/serialization/MapCodec;
+			ARG 0 type
+		METHOD m_pfcndfzi createOverride (Lnet/minecraft/unmapped/C_rdaqiwdt;)Lnet/minecraft/unmapped/C_azfkqhlm$C_kqgfvdat;
+		METHOD m_tfwgfjdw getType ()Lnet/minecraft/unmapped/C_azfkqhlm$C_kqgfvdat$C_ojlwbbdy;
+		METHOD m_vdlizjqr getHidden ()Lnet/minecraft/unmapped/C_azfkqhlm$C_kqgfvdat;
+		CLASS C_dzmzhicx Default
+			FIELD f_bnylsies INSTANCE Lnet/minecraft/unmapped/C_azfkqhlm$C_kqgfvdat$C_dzmzhicx;
+		CLASS C_mgthfrsm TextOverride
```
supersaiyansubtlety, line 44: I initially named this `Override`, but after I rebased to 25w18a, that started causing errors for javadoc:
> error: incompatible types: Override cannot be converted to Annotation

It was getting it confused with `java.lang.Override`.

If there's a way we can ensure the fake source gets generated with qualified `java.lang.Override`s in cases like this that would be ideal.

mappings/net/minecraft/entity/passive/AgingEntity.mapping
```diff
@@ -1,8 +1,10 @@
-CLASS net/minecraft/unmapped/C_jvojbnla net/minecraft/entity/passive/PassiveEntity
+CLASS net/minecraft/unmapped/C_jvojbnla net/minecraft/entity/passive/AgingEntity
```
supersaiyansubtlety, line 1: everything in the class has to do with aging

mappings/net/minecraft/entity/passive/AgingEntity.mapping
```diff
@@ -20,8 +22,8 @@ CLASS net/minecraft/unmapped/C_jvojbnla net/minecraft/entity/passive/PassiveEnti
 	METHOD m_rrykeatb setBreedingAge (I)V
 		ARG 1 age
 	METHOD m_tjkaggcz getRemainingHappyTicks ()I
-	METHOD m_xdpbetzz onGrowUp ()V
-	CLASS C_uxhqdxwl PassiveData
+	METHOD m_xdpbetzz onAgeStageChange ()V
```
supersaiyansubtlety, line 25: this also handles "ungrowing"

mappings/net/minecraft/entity/passive/TamableEntity.mapping
```diff
@@ -1,11 +1,12 @@
-CLASS net/minecraft/unmapped/C_sjwjnzbw net/minecraft/entity/passive/TameableEntity
+CLASS net/minecraft/unmapped/C_sjwjnzbw net/minecraft/entity/passive/TamableEntity
```
supersaiyansubtlety, line 1: afaict both spellings are valid
I'm fine with either spelling
`TamableMountEntity` already used this spelling and it's one fewer character

mappings/net/minecraft/text/component/NbtTextComponent.mapping
```diff
@@ -1,4 +1,4 @@
-CLASS net/minecraft/unmapped/C_tvzpgqlc net/minecraft/text/component/NbtComponent
+CLASS net/minecraft/unmapped/C_tvzpgqlc net/minecraft/text/component/NbtTextComponent
```
supersaiyansubtlety, line 1: to eliminate confusion with `n.m.component.type.NbtComponent`

mappings/net/minecraft/data/DataWriter.mapping
supersaiyansubtlety, line 1: renamed -> `DataCache`

mappings/net/minecraft/entity/Entity.mapping
```diff
@@ -716,6 +800,8 @@ CLASS net/minecraft/unmapped/C_astfners net/minecraft/entity/Entity
 		ARG 6 z
 		ARG 8 yaw
 		ARG 10 pitch
+	METHOD m_tggkpsnn tryShearLeashes (Lnet/minecraft/unmapped/C_jzrpycqo;)Z
+	METHOD m_tgqrwdfb isMainPlayerControlled ()Z
```
eristhea, line 804: this could use a javadoc clarification to explain that 'main player' means the player on the current client

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
```diff
@@ -52,7 +56,9 @@ CLASS net/minecraft/unmapped/C_tahhkgty net/minecraft/entity/vehicle/AbstractMin
 		ARG 1 offset
 	METHOD m_xtdpuoft willHitBlockAt (Lnet/minecraft/unmapped/C_hynzadkk;)Z
 		ARG 1 pos
+	METHOD m_xzvoogyl getContainedBlockOrEmpty ()Ljava/util/Optional;
```
eristhea, line 59: noticing an inconsistency here. not an issue for this PR, but i'm not sure whether we should prefer `OrEmpty` or `Optional` as a suffix for optional values. I personally prefer optional, but I think I would ascribe that to experience and a newer programmer would appreciate OrEmpty. I'll open an issue!

eristhea (approved):
amazing work! I feel like I have less and less to say with each review on your PRs :)

mappings/net/minecraft/entity/vehicle/AbstractMinecartEntity.mapping
supersaiyansubtlety: This is related to #665
I could just broaden that to all getters, unless you'd prefer a separate issue.

I prefer `OrEmpty` because it reads more naturally. `getContainedBlockOptional` isn't natural at all; `getContainedOptionalBlock` is better, but it's not as simple as a suffix, and imo it still isn't as natural as `OrEmpty`.

eristhea: i'm out of touch after being gone for so long :sob: 
i'm with you on `OrEmpty`!

PR #703, opened 2025-05-03
complete `n.m.village`

supersaiyansubtlety:
also maps `AllComponentsPredicate` and `SavedDataType`


mappings/net/minecraft/village/raid/Raid.mapping
```diff
@@ -1,5 +1,5 @@
 CLASS net/minecraft/unmapped/C_szefbyex net/minecraft/village/raid/Raid
-	FIELD f_apbpyjyx waveToRaiders Ljava/util/Map;
```
eristhea, line 2: straight up bizarre old name. what were we cooking

eristhea (approved):
thanks :)

PR #704, opened 2025-05-03
Complete `n.m.util`

supersaiyansubtlety:
(mostly) completes `n.m.util`
(a few params and fields in `math` are weird)


mappings/net/minecraft/util/math/Axis.mapping
supersaiyansubtlety, line 1: renamed -> `AxisRotation`

OroArmor (approved):

PR #713, opened 2025-06-23
Restore equipment renderer

Pixaurora:
This just merges in Saiyan's QSL mappings pr. I might've(?) missed something else, but I'm not sure


supersaiyansubtlety (approved):
oops forgot to port this, thanks!

PR #725, opened 2025-07-21
Complete `n.m.entity` 1.21.8

supersaiyansubtlety:
- renames `Task` -> `ContinuableTask`
- renames `TaskControl` -> `Task`
- renames methods that return passed `Brain`s `create` -> `init`
- adds several simple type names
- *removes* `inherit` from `Waypoint` simple name because it's inherited by entities (I added new simple names for appropriate subclasses)


supersaiyansubtlety:
I'm keeping this on 1.21.8 so it can be on a release version sooner.
I'll port to the latest version once it's merged.

PR #735, opened 2025-09-04
Update QEP

supersaiyansubtlety:
Updates QEP to 2.4.0 to fix this: <https://github.com/QuiltMC/quilt-mappings/actions/runs/17413250613/job/49435424575#step:8:1299>

Also uses the static `run` method for `dropInvalidMappings`; it's cleaner and it will need to be changed for enigma 2.7 anyways.


eristhea (approved):
looks great!

PR #737, opened 2025-10-07
Added mappings for Creaking Home Pos

HyperPigeon:
- Added a small handful of mappings related to the creaking's home position.


supersaiyansubtlety:
These are covered by [#725](https://github.com/QuiltMC/quilt-mappings/pull/725/files#diff-3118f4e0a7c441d99f8a8826774a83a1d8ebb451b33ee83022afe8db9217dee9) (which will be ported forward once merged).

Sorry, I know it's difficult to check for overlap with other PRs when there are quite a few and some are large. We've been short on reviews lately.

PR #740, opened 2025-10-12
 added mappings for previously unmapped game mode list

HyperPigeon:


mappings/net/minecraft/unmapped/GameModeList.mapping
```diff
@@ -0,0 +1,5 @@
+CLASS net/minecraft/unmapped/C_uafqtlvc net/minecraft/unmapped/GameModeList
+	FIELD f_rtbplhce ALL_GAME_MODES Lnet/minecraft/unmapped/C_uafqtlvc;
+	METHOD m_eaoscrej containsGameMode (Lnet/minecraft/unmapped/C_lghcpyvl;)Z
```
supersaiyansubtlety:
```suggestion
	METHOD m_eaoscrej contains (Lnet/minecraft/unmapped/C_lghcpyvl;)Z
```
Since the class is all about game modes, I don't think there's a need to specify the type here; better to keep it brief.

supersaiyansubtlety (changes requested):
Looks great, thanks!

Just needs to go in a package; I suggest `net/minecraft/predicate` since that's were it's used.

supersaiyansubtlety (approved):
Looks good!

PR #743, opened 2025-10-20
Fix build number

supersaiyansubtlety:
This should fix what #742 tried and failed to fix.
#742 didn't work because the project version changed between the `build javadocJar` and `publish` invocations, so `build javadocJar`'s outputs where out-of-date, so they re-ran before `publish`.

This:
- re-combines `build javadocJar` with `publish`
- initializes the `BUILD_NUMBER` env variable (starting at ~~0~~1) before "Publish"
- increments the build number after "Publish"
- ~~makes `build.gradle` add 1 to `BUILD_NUMBER` so it doesn't start at 0~~ the workflow adds 1 now

I "tested" on this repo which just has fake publish tasks and workflows:
https://github.com/supersaiyansubtlety/publish-build-test

This branch's action uses a fake publish task that always fails:
https://github.com/supersaiyansubtlety/publish-build-test/tree/publish-fail


.github/workflows/publish.yml
```diff
@@ -21,18 +26,23 @@ jobs:
...
+      # Fetch tags because they're used to store build number.
+      - name: Fetch tags
+        run: git fetch --tags origin
+      # Check for existing tag created by previous runs of "Increment build number".
+      # Extract build number if present, otherwise start at 0.
+      - name: Set build number
+        shell: bash
+        run: |
+          output=$(git tag -l build/${{ github.ref }}-build-number-*)
+          buildNumber=$(if [[ $output ]]; then echo $output | grep -oP "(?<=build-number-)[0-9]+$"; else echo 0; fi)
```
supersaiyansubtlety: Note that this assumes only one tag was found; this should be true so long as we don't set `delete_previous_tag: false` for `onyxmueller/build-tag-number@v1`.

.github/workflows/publish.yml
```diff
@@ -8,10 +8,15 @@ jobs:
       image: eclipse-temurin:21
       options: --user root
     steps:
+      - name: Install git
+        run: apt-get update && apt-get install -y git
+      - name: Configure git for container
+        run: git config --global --add safe.directory "$(pwd)"
+
       - uses: actions/checkout@v4
```
IotaBread, line 16: I think you can skip installing, configuring git and fetching the tags manually by using the [`fetch-tags: true` option for `actions/checkout`](https://github.com/actions/checkout/tree/v4?tab=readme-ov-file#usage) (though maybe the install step is required to get the build number, but the readme for the action does show [an example where its not needed](https://github.com/actions/checkout/tree/v4?tab=readme-ov-file#usage), maybe it has to do with the container image we're using)

supersaiyansubtlety: I tried `fetch-tags: true` on the test repo and it didn't work:
<https://github.com/supersaiyansubtlety/publish-build-test/actions/runs/18693488525/job/53304836878#step:10:22>

The `fakePublish` task had build number `1` despite the existing `build-number-N` tag because "Set build number" found no tags and defaulted to `0`.

I think it is because of the container.

IotaBread: oh ok

supersaiyansubtlety:
superseded by #744

